PR #42, opened 2023-04-17
Compare models

pseudotensor:
When click checkbox for compare, sections open up and new mode occurs

- [x] ~Auto-unload model2 when deselect compare models~ won't do, expert control
- [x] Give option to load and unload each in turn in case memory  hungry, but can do manually currently.
- [x] name chatbot with model name instead of h2oGPT2 etc.
- [x] When add new lora, don't switch dropdown lora if already loaded model for that model

image

image

image


pseudotensor:
Now support choosing specific options for model loading for controlling comparison:

image

image

pseudotensor:
But if that other GPU OOMs, model.generate() spits back error as response, and regenerate gets funky:

image

PR #46, opened 2023-04-19
Add code to push spaces chatbot

pseudotensor:
generate works still
image


arnocandel (approved):

arnocandel (commented):
https://github.com/h2oai/h2ogpt/blob/9ee2205072cf1aa178a37c2346e671bd7106c779/export_hf_checkpoint.py#L178 ~needs to update path after moving the h2oai_pipeline.py~ was just moving the link, all good, tx

PR #60, opened 2023-04-20
command fix in install.md

fatihozturkh2o:
This needs to be h2ogpt, not h2o-llm?


arnocandel:
thanks, we just merged another PR for the same.

PR #68, opened 2023-04-21
h2o theme for background

jefffohl:
This PR removes the repeating H2O logo from the background, and - hopefully - incorporates the H2O color theme to the app. I say hopefully, because I have not yet got a local dev environment running, lacking CUDA, and probably some other things necessary to successfully run `generate.py` - so I was not able to test this locally.

This is just a simple fix - but it might be good to start working more on a customized UI of some sort, if that is warranted at this point.

The UI should look something similar to this (though with slightly different colors for the yellows and grays):

Light mode:
Screenshot 2023-04-20 at 6 23 43 PM

Dark mode:
Screenshot 2023-04-20 at 6 15 29 PM

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -544,7 +544,7 @@ def go_gradio(**kwargs):
         task_info_md = ''
 
     css_code = """footer {visibility: hidden;}
-body{background:linear-gradient(#f2f2f2,#cccccc);}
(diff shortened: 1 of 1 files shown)


arnocandel (approved):
Thanks!

PR #71, opened 2023-04-21
Use SAVE_DIR and --save_dir instead of SAVE_PATH and --save_path.

arnocandel:
Now must be directory. File will be called history.json.


pseudotensor (approved):
tested and works

PR #74, opened 2023-04-22
Make chat optional from UI/client

pseudotensor:
And simplify client to just chat=False case since chat requires going through disk jsons and not useful/required for client calls.

### DIFF
--- a/client_test.py
+++ b/client_test.py
@@ -1,9 +1,9 @@
 """
-Client test.  Simplest case is chat=False and stream_output=False
+Client test.
 
-Run server with same choices:
+Run server:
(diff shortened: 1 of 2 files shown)


pseudotensor:
checkbox in expert can switch chat mode to simple non-chat mode:

image


Both chat and non-chat tracks are entirely independent, and one can switch to the other in the UI without losing the other case:

image


For client code, it should always take the chat=False track, because the gradio chat goes through tile system with json files to do its work, which isn't available remotely.  Also too hard to maintain that track due to how inputs/outputs processed for chatbot.

PR #85, opened 2023-04-26
Refactor gradio into separate file and isolate it from torch specific stuff

pseudotensor:
Runs fine:

image


pseudotensor:
Still works

pseudotensor:
Still works:

image


image

arnocandel (approved):

pseudotensor:
All tests pass, user tests pass, so will assume ok.  use as way to add new tests if something ends up failing.

PR #93, opened 2023-04-27
Refactor finetune so some of it can be used to check data and its tokenization

pseudotensor:
image


pseudotensor:
Token dist with 512 cutoff:

image

PR #98, opened 2023-04-29
Try transformers experimental streaming.  Still uses threads, so probably won't fix browser exit GPU memory issue

pseudotensor:
One this is fixed:  https://github.com/gradio-app/gradio/issues/4016 then can incorporate that and rely upon stopping to clear activate generate instead of just waiting for those threads to finish.


pseudotensor:
High concurrency (10) runs fine (just slower), no more memory.

image


and hit this:

https://github.com/h2oai/h2ogpt/issues/104

this breaks everything.

PR #117, opened 2023-05-06
Reform chatbot client API code

pseudotensor:
but serialize=False doesn't work due to gr.state issue: https://github.com/gradio-app/gradio/issues/4081

Original here that had to use json files due to serialize=True default: https://github.com/h2oai/h2ogpt/blob/05c742a1a283bf307bc745e8f276b0678f0ed9a9/client_test.py#L81-L107


pseudotensor:
```

==================================================================================== 4 passed, 1 skipped in 310.72s (0:05:10) =====================================================================================
(h2ollm) jon@pseudotensor:~/h2o-llm$ 

```

PR #131, opened 2023-05-15
[DevOps] Snyk Integration

ChathurindaRanasinghe:
- Scan results can be found [here](https://app.snyk.io/org/h2o-gpt/projects).
- `requirements.txt` is scanned for vulnerabilities & license issues.
- Following requirements are removed from scan.
https://github.com/h2oai/h2ogpt/blob/2ccf687ea3f3f3775a501838e81c1a0066430455/requirements.txt#L42-L51

Parent Issue: https://github.com/h2oai/h2o-ops/issues/323

### DIFF
--- a/.github/workflows/snyk-scan.yml
+++ b/.github/workflows/snyk-scan.yml
@@ -20,8 +20,16 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '3.10'
+      
+      - name: Check changed Deps files
+        uses: tj-actions/changed-files@v35
+        id: changed-files
+        with:
+          files: | # This will match all the files with below patterns
+            requirements.txt
     
       - name: Scan python dependencies
+        if: contains(steps.changed-files.outputs.all_changed_and_modified_files, 'requirements.txt')
         env:
           SNYK_TOKEN: '${{ secrets.SNYK_TOKEN }}'
         run: |
--- a/.github/workflows/snyk-scan.yml
+++ b/.github/workflows/snyk-scan.yml
@@ -1,0 +1,64 @@
+name: Snyk Security Vulnerability Scan
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+    branches:
+      - main
+      - devops/snyk-integration
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/snyk-scan.yml
```diff
@@ -0,0 +1,63 @@
+name: Snyk Security Vulnerability Scan
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
```
hasithjp, line 8: I don't see any tag here. please verify with the code owner

hasithjp (approved):
LGTM!. thanks

.github/workflows/snyk-scan.yml
ChathurindaRanasinghe: That was added since there were no release branches or tags and assuming in future there may be releases.
@arnocandel can you tell us how the releases will be done?

pseudotensor: It's great point, we really need to make pypi package release.  It would have been a bit premature before, as I even noticed e.g. pandoc would have been trouble for some because really need pandoc binary.  But I tested things on HF and other platforms, so it should be quite compatible with varying environments.

pseudotensor (approved):
How do I evaluate the goodness of this PR?  Does it hook-up to our internal devops scans?  Can those be made public with some link in the readme?

ChathurindaRanasinghe:
> How do I evaluate the goodness of this PR? Does it hook-up to our internal devops scans? Can those be made public with some link in the readme?

If there are any issues with PR the checks will fail. You can check the logs for GitHub action workflows. Those are publicly available.

pascal-pfeiffer:
> You can check the logs for GitHub action workflows. Those are publicly available.

They are not afaik. The logs can (and do, I believe) contain secret tokens printed to stdout. It makes sense to keep those logs private to the repo maintainers.

ChathurindaRanasinghe:
> > You can check the logs for GitHub action workflows. Those are publicly available.
> 
> They are not afaik. The logs can (and do, I believe) contain secret tokens printed to stdout. It makes sense to keep those logs private to the repo maintainers.

sure

arnocandel (approved):

PR #133, opened 2023-05-15
Add h2oGPT Client

### LINKED ISSUE #80: Create API wrapper so like OpenAI's API
https://platform.openai.com/docs/guides/chat/introduction

https://github.com/openai/openai-cookbook/blob/main/examples/How_to_format_inputs_to_ChatGPT_models.ipynb

Then our non-chat and chat cases would also be covered, but as API at python level instead of bare bones API calls with api_name strings.

fastchat did this, apache2, so can copy: https://github.com/lm-sys/FastChat/blob/main/fastchat/client/test_client.py

this:
Fixes #80


pseudotensor:
@this Looking great!  Make sure you merge main into your PR as the langchain had breaking changes with 1 extra API item.

Also, gradio team fixed the Chat part.  I'll update that now:

https://github.com/gradio-app/gradio/issues/4081#issuecomment-1548985863

client/h2ogpt_client/core.py
```diff
@@ -0,0 +1,208 @@
...
+    QUALITY = "quality"
+    HUMAN_BOT = "human_bot"
+    DAI_FAQ = "dai_faq"
+    SUMMARIZE = "summarize"
+    SIMPLE_INSTRUCT = "simple_instruct"
+    INSTRUCT_VICUNA = "instruct_vicuna"
+    INSTRUCT_WITH_END = "instruct_with_end"
+    HUMAN_BOT_ORIGINAL = "human_bot_orig"
+    PROMPT_ANSWER = "prompt_answer"
+    OPEN_ASSISTANT = "open_assistant"
```
arnocandel, line 20: https://github.com/h2oai/h2ogpt/blob/c18a4282e51119eab44465319d0f563756bceb9b/prompter.py#L18 we now have more, no better way to keep in sync? maybe using Makefile to auto-extract?

pseudotensor: He can move that prompter.py enum to a new file that can be consumed by client and server.  There are a few things like that, the client needs to know things.

However, often, the model knows what prompt it needs, so should be able to support passing '' typically and it would just use the right default if exists.

Or, if we used pipelines, then even less needs to be known.

pseudotensor:
@this How's this PR going?

client/h2ogpt_client/core.py
this: Rather than specifying a `PromptType`, can we give the option to specify a model name? e.g. `client.text_completion.create("How are you?", model="distilgpt2")`. OpenAI API works likes this

client/h2ogpt_client/core.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        # Not exposed parameters.
+        you = ""  # empty when chat_mode is False
+        stream_output = False
+        chat_mode = False
+
+        return await self._client._predict_async(
+            you,
+            input,
+            system_pre_context,
```
this, line 189: @pseudotensor If the user has given a non-empty `prompt`, then setting the `system_pre_context` make sense?

this:
> How's this PR going?

@pseudotensor I'm currently working on the `chat_completion` API. I was hoping to include that in this PR, but we can merge this and I can create a new PR.

client/README.md
```diff
@@ -0,0 +1,21 @@
+# h2oGPT Client
+A Python thin-client for h2oGPT.
```
pseudotensor, line 2: Add note that we have OpenAI-compliant API here: https://github.com/h2oai/h2ogpt#client-api

client/h2ogpt_client/core.py
```diff
@@ -0,0 +1,348 @@
+import asyncio
+import collections
+from enum import Enum
+from typing import Any, Dict, List, Optional, OrderedDict, Tuple
+
+import gradio_client  # type: ignore
+
+
+class PromptType(Enum):
```
pseudotensor, line 9: Yes please import from prompter.py instead.

client/h2ogpt_client/core.py
pseudotensor: For now model is fixed from client side, but I'll fix that separately.  For now, user shouldn't have to specify anything about model or prompt_type, although they should be able to specify prompt_type etc. if they want to override defaults.

client/h2ogpt_client/core.py
```diff
@@ -0,0 +1,348 @@
...
+    PLAIN = "plain"
+    PROMPT_ANSWER = "prompt_answer"
+    QUALITY = "quality"
+    SIMPLE_INSTRUCT = "simple_instruct"
+    SUMMARIZE = "summarize"
+    WIZARD_LM = "wizard_lm"
+    WIZARD_MEGA = "wizard_mega"
+
+
+class LangChainMode(Enum):
```
pseudotensor, line 29: I changed order and put in a new file so can be used in general.  Need order for UI etc.: https://github.com/h2oai/h2ogpt/pull/260

client/h2ogpt_client/core.py
```diff
@@ -0,0 +1,348 @@
...
+
+    DAI_FAQ = "dai_faq"
+    HUMAN_BOT = "human_bot"
+    HUMAN_BOT_ORIGINAL = "human_bot_orig"
+    INSTRUCT = "instruct"
+    INSTRUCT_SIMPLE = "instruct_simple"
+    INSTRUCT_VICUNA = "instruct_vicuna"
+    INSTRUCT_WITH_END = "instruct_with_end"
+    OPEN_ASSISTANT = "open_assistant"
+    PLAIN = "plain"
```
pseudotensor, line 20: We should avoid dup with rest of code as this is constantly extended.  I moved to enum the original version.  Is there some reason to stick to FOO_BAR = "foo_bar" format?

client/README.md
this: Added in 0bada300e758b73e48952f242ec70a37c1b2358a

client/h2ogpt_client/core.py
this: > Is there some reason to stick to FOO_BAR = "foo_bar" format?

@pseudotensor I simply capitalized the value to derive the name of the enum. Id these names are not appropriate, please suggest good names.

All capitalized names were used by following the Python convention https://docs.python.org/3.8/library/enum.html#creating-an-enum. I noticed that [`PromptType` enum names](PromptType) are all simple. Shall we change that?

client/h2ogpt_client/core.py
pseudotensor: yes, it's correct.

client/h2ogpt_client/core.py
```diff
@@ -0,0 +1,314 @@
...
+        top_p: float = 1.0,
+        top_k: int = 40,
+        beams: float = 1.0,
+        early_stopping: bool = False,
+        min_output_length: int = 0,
+        max_output_length: int = 128,
+        max_time: int = 180,
+        repetition_penalty: float = 1.07,
+        number_returns: int = 1,
+        system_pre_context: str = "",
```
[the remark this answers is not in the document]
pseudotensor: E.g. you can at build time check in python that set of inspection arguments for "create" is larger than set from `eval_func_param_names` .  I do stuff like that in h2oGPT.

client/h2ogpt_client/core.py
pseudotensor: Yes, let's change to whatever is normal convention.  Requires going through and changing conditionals where PromptType used.

client/h2ogpt_client/core.py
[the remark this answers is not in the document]
this: ya, we can do that. Shall we create a new PR for this? In order to make `eval_func_param_names` isolated (so that I can be used in the client) I had to make changes in places where `eval_func_param_names` is used. I fell like if I add those changes into this PR, it will clutter the PR.

client/h2ogpt_client/core.py
this: I'll create a separate PR for this.

pseudotensor (approved):
Let's merge!  Thanks!  Let's do more testing and refactors in another PR.

pseudotensor:
Non-manual test failures on this branch:

```
============================================================================================================ short test summary info ============================================================================================================
FAILED tests/test_pipeline.py::test_export_copy - ImportError: cannot import name 'inject_newline' from 'test_output.h2oai_pipeline' (/home/jon/h2ogpt/test_output/h2oai_pipeline.py)
FAILED tests/test_requirements.py::test_requirements - ValueError: ['accelerate==0.19.0', 'transformers==4.28.1', 'git+https://github.com/huggingface/peft.git@3714aa2fff158fdfa637b2b65952580801d890b2']
FAILED tests/test_ui.py::test_newline_replace - AssertionError: assert 'You can use ... sorted list.' == 'You can use ... sorted list.'
=========================================================================== 15 failed, 99 passed, 36 skipped, 1 xfailed, 1 xpassed, 12 warnings in 3745.08s (1:02:25) ===========================================================================
(h2ollm) jon@pseudotensor:~/h2ogpt$ 
```

Unrelated to this PR.

PR #154, opened 2023-05-21
fixes for upload gui

jefffohl:
This PR makes some adjustments to the GUI:

- Makes the text in the upload interface more legible in dark mode
- Adds file types to the upload interface
- Tightens up the overall GUI with some adjustments for text size, font size, and margins, and rounded corners, to make the UI more usable on smaller screens

BEFORE:
Screenshot 2023-05-21 at 1 24 45 PM

AFTER:
Screenshot 2023-05-21 at 1 23 01 PM


gradio_runner.py
```diff
@@ -271,19 +268,18 @@ def upload_file(files, x):
                         from gpt_langchain import file_types, have_arxiv
                     else:
                         have_arxiv = False
-                        file_types = []
+                        file_types = ['pdf', 'txt', 'csv', 'toml', 'py', 'rst', 'rtf', 'md', 'html', 'enex', 'eml', 'epub', 'odt', 'pptx', 'ppt', 'zip', 'urls', 'docx', 'doc', 'png', 'jpg', 'jpeg']
```
pseudotensor, line 271: These are not available if  langchain is disabled, why was [].  Just dummy value to avoid importing langchain when disabled.

jefffohl: Ah, OK - so we would want some conditional logic then, right? But, I suspect there is something I am not understanding.

pseudotensor:
Colors are a bit dulled out.  Makes it a bit harder to identify structure of things.

jefffohl:
Hey @pseudotensor - apologies for the extreme lateness of this response. Would you like to see another iteration?

pseudotensor:
Hi @jefffohl I think we can close this PR, I think the solution used in main is ok and minimal.

Issue #136, opened 2023-05-16
How come no 4bit?


bbecausereasonss:
Curious why no 4bit models. Would really amplify the ability to use larger models.

arnocandel:
https://github.com/h2oai/h2ogpt/pull/107 we are working on 4-bit quantization

chris-aeviator:
is 4 bit usable with the latest merge?

pseudotensor:
@chris-aeviator , for GPU models on newer GPUs, the latest bitsandbytes supports 4-bit yes.

See the middle part of readme.md for GPU section: https://github.com/h2oai/h2ogpt#gpu-cuda

pseudotensor:
bitsbytes supplorts

PR #107, opened 2023-05-02
Add Quantization code.

arnocandel:
Add quantization logic #88

PR #166, opened 2023-05-25
Use latest peft/transformers/accelerate/bitsandbytes for 4-bit (qlora)

arnocandel:
https://arxiv.org/abs/2305.14314
https://huggingface.co/blog/4bit-transformers-bitsandbytes
- [x] update deps
- [x] enable 4-bit generation
- [x] enable 4-bit lora training

Addresses #136 more directly than #107, but only at runtime.


finetune.py
```diff
@@ -65,7 +65,8 @@ def train(
         micro_batch_size: int = 4,
         gradient_checkpointing=False,  # unnecessary with gradient accumulation enabled
         fp16=True,
-        train_8bit=True,
+        train_8bit=False,
```
arnocandel, line 68: disable 8-bit training by default, since much slower than 16-bit.

generate.py
```diff
@@ -270,10 +271,12 @@ def main(
                 base_model = 'h2oai/h2ogpt-oasst1-512-12b'
                 # don't set load_8bit if passed base_model, doesn't always work so can't just override
                 load_8bit = True
+                load_4bit = False  # FIXME - consider using 4-bit instead of 8-bit
```
arnocandel, line 274: first need to do performance testing, for both speed & accuracy

arnocandel:
Testing 65B on 2x A6000Ada:
`(env) arno@rippa:/nfs4/llm/h2ogpt(4bit)$ CUDA_VISIBLE_DEVICES=0,1 torchrun --nproc_per_node=2 finetune.py --base_model=decapoda-research/llama-65b-hf --train_4bit=True --micro_batch_size=1 --run_id=3 --data_path=h2oai/openassistant_oasst1_h2ogpt_graded`
works:
`  0%|▋                                                                                                                                                            | 1/237 [04:27<17:33:37, 267.87s/it]`
image

pseudotensor (approved):
We will do more full testing of eval on 4bit vs. 8bit vs. 16bit.  According to QLORA paper, as long as do LORA on all linear layers, then should be as good as 16bit.  I'm unclear on speed of new Tim stuff.

PR #230, opened 2023-06-02
Use optional dependencies

achraf-mer:


README.md
```diff
@@ -316,21 +316,29 @@ We disable background uploads by disabling telemetry for huggingface, gradio, an
 
 #### Python Wheel
 
-The wheel adds all dependencies except specialized dependencies like 4-bit and flash-attention.  For GPU build do:
+The wheel adds all dependencies including optional dependencies like 4-bit and flash-attention. To build do:
```
pseudotensor, line 319: These shouldn't be included.

requirements.txt
```diff
@@ -18,9 +18,9 @@ pandas==2.0.0
 matplotlib==3.7.1
 loralib==0.1.1
 bitsandbytes==0.39.0
-accelerate==0.19.0
+accelerate>=0.19.0
 git+https://github.com/huggingface/peft.git@3714aa2fff158fdfa637b2b65952580801d890b2
-transformers==4.28.1
+transformers>=4.28.1
```
pseudotensor, line 23: Can't do this.

setup.py
```diff
@@ -1,99 +1,81 @@
...
     install_requires=install_requires,
-    # https://pypi.org/classifiers/
+    extras_require={
+        'ALL': install_gpl + install_extra_training + install_wiki_extra,
+        'CPU': install_no_gpu,
+        'GPU': install_gpu,
+        '4BIT': install_4bit,
+        'FLASH': install_flash,
+        'GPL': install_gpl,
+        'TRAINING': install_extra_training,
```
pseudotensor, line 67: I can't follow why this is done, can't require all deps.

setup.py
```diff
@@ -1,99 +1,81 @@
...
     with open(file_name) as f:
         required = f.read().splitlines()
     required = [x for x in required if not x.strip().startswith("#")]
-    required = [x if 'git+http' not in x else 'peft @' + x for x in required]
+    required = [x if 'git+http' not in x else re.search(r"/([^/]+?)\.git", x).group(1) + ' @ ' + x for x in required]
     required = [x for x in required if x]
     return required
 
 
-do_install_optional_req = bool(int(os.environ.get('OPTIONAL', '1')))
```
pseudotensor, line 18: I'm unclear on why this is all necessary, seems like arbitrary refactor, and assumes wheel should always use all optional things, which is not right.

README.md
achraf-mer: all dependencies are defined, but won't be used/installed unless explicitly instructed by user using the `[..]` suffix.

setup.py
achraf-mer: just holds the definition for the sub-dependencies, the defaults are still the same (`requirements.txt`), the dict above is just for optionals

setup.py
achraf-mer: not arbitrary, we discussed in slack, should help to have one single whl so can deploy to pypi.

achraf-mer:
using this instead: https://github.com/h2oai/h2ogpt/pull/243

PR #243, opened 2023-06-06
Add ci support + wheel

achraf-mer:


pseudotensor (approved):
Few questions asked offline, plus maybe more examples for different use cases good to have.

pseudotensor:
The conflict is because we had to update reqs.  Now, lucikly, no need for 8bit vs. 4bit as new transformers works from pypi for both properly, all tests pass including 4-bit ones.
So you can remove that part of the PR.  For req_constraints.txt, use main, or just rebuild with make

arnocandel (approved):

PR #257, opened 2023-06-08
venv support in gitignore&Makefile

fatihozturkh2o:
As someone actively using the h2ogpt repo, I create my Python venv in the project folder before running the models. However, at every git pull, I need to delete venv because can not pull without stashing changes for the venv.

I think there could be many others running into the same issue, so thought this can be a nice slight improvement :)

(Also added helpers in Makefile, but your call)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,7 @@ flagged_data_points
 .pytest_cache
 user_path
 user_path_test
+
+# Environments
+venv/
+.venv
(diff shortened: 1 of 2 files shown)


pseudotensor (commented):
Similar changes in https://github.com/h2oai/h2ogpt/pull/133/files

fatihozturkh2o:
> Similar changes in https://github.com/h2oai/h2ogpt/pull/133/files

great 👍

PR #258, opened 2023-06-08
prompt form moved to a separate file

fazpu:
- new `gradio_ui` directory was created with a first file `prompt_form` to which I extracted the prompt form logic
  - I believe this will improve code structure and readability in long run

### DIFF
--- a/gradio_ui/prompt_form.py
+++ b/gradio_ui/prompt_form.py
@@ -1,0 +1,21 @@
+import gradio as gr
+
+
+def make_prompt_form(kwargs):
+    if kwargs['input_lines'] > 1:
+        instruction_label = "You (Shift-Enter or push Submit to send message, use Enter for multiple input lines)"
+    else:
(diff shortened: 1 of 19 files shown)


pseudotensor:
Hi @fazpu , it's not what ChatGPT does or HuggingChat does, etc.  It feels more normal original way.  But happy to discuss.

fazpu:
> Hi @fazpu , it's not what ChatGPT does or HuggingChat does, etc. It feels more normal original way. But happy to discuss.

@pseudotensor got it. Only the code refactoring changes remained.

gradio_ui/prompt_form.py
```diff
@@ -0,0 +1,21 @@
+import gradio as gr
```
pseudotensor, line 1: Have to be careful with new files.  Need to then change repo_to_spaces.sh

pseudotensor (approved):

PR #259, opened 2023-06-08
input prompt - primary color border added + change in label text

fazpu:
continuation of https://github.com/h2oai/h2ogpt/pull/258

- input prompt - primary-color border added + change in label text
- css code moved to separate file

## h2ocolors=True
image

image

## h2ocolors=False
image

image

### DIFF
--- a/gradio_ui/css.py
+++ b/gradio_ui/css.py
@@ -1,0 +1,38 @@
+def get_css(kwargs) -> str:
+    if kwargs['h2ocolors']:
+        css_code = """footer {visibility: hidden;}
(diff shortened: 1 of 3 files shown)


gradio_ui/css.py
```diff
@@ -0,0 +1,38 @@
+def get_css(kwargs) -> str:
```
pseudotensor, line 1: Same issue as other PR, need to add new files to repo_to_spaces.sh

pseudotensor: will do separately

PR #274, opened 2023-06-10
A couple of small updates to the documentation

3x0dv5:
Describe a few dependencies that might need to be installed before installing the requirements.
`tesseract-ocr` was present twice in the `apt install`


3x0dv5:
@pseudotensor a small PR, could you take a peek please?

pseudotensor (approved):
Thanks!

PR #276, opened 2023-06-11
Escape so it outputs `$LD_LIBRARY_PATH:/usr/local/cuda/lib64/`

3x0dv5:
Escape so it outputs `$LD_LIBRARY_PATH:/usr/local/cuda/lib64/` and not `{value of $LD_LIBRARY_PATH}:/usr/local/cuda/lib64/`

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -49,10 +49,9 @@ sudo apt-get -y install cuda
 
 Then set the system up to use the freshly installed CUDA location:
 ```bash
-echo "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda/lib64/" >> ~/.bashrc
-echo "CUDA_HOME=/usr/local/cuda" >> ~/.bashrc
-echo "export PATH=$PATH:/usr/local/cuda/bin/" >> ~/.bashrc
-echo "conda activate h2ogpt" >> ~/.bashrc
+echo "export LD_LIBRARY_PATH=\$LD_LIBRARY_PATH:/usr/local/cuda/lib64/" >> ~/.bashrc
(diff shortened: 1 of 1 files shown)


3x0dv5:
@pseudotensor I come across a small issue in the doc, because we are not escaping the string we are outputting the value of the variable into the bashrc and not the ${Variable}. 
Changed the docs, if you think that this is the right way, please merge.

docs/INSTALL.md
```diff
@@ -49,9 +49,9 @@ sudo apt-get -y install cuda
 
 Then set the system up to use the freshly installed CUDA location:
 ```bash
-echo "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda/lib64/" >> ~/.bashrc
+echo "export LD_LIBRARY_PATH=\$LD_LIBRARY_PATH:/usr/local/cuda/lib64/" >> ~/.bashrc
 echo "CUDA_HOME=/usr/local/cuda" >> ~/.bashrc
-echo "export PATH=$PATH:/usr/local/cuda/bin/" >> ~/.bashrc
+echo "export PATH=\$PATH:/usr/local/cuda/bin/" >> ~/.bashrc
 echo "conda activate h2ogpt" >> ~/.bashrc
```
3x0dv5, line 55: @pseudotensor do we really need this `conda activate` to in the bashrc?

pseudotensor: Please remove it, it is already there externally a couple lines lower. Thanks!

pseudotensor (changes requested):
Please remove the echo of conda activate into the .bashrc and then I'll approve, thanks!

3x0dv5:
Done, thanks

pseudotensor (approved):
thanks!

PR #328, opened 2023-06-24
Add explicit cast to bool for visible=kwarg[...]

parkeraddison:
Based on the Svelte source, Gradio blocks are only hidden when visible=False. Therefore certain arguments with default values of None cause an unintended effect of displaying a block when visible=None (when the intention is to be falsey). Example: the warning message for `model_lock` showed up even when it was unspecified (None).

So as to make minimal modifications to the code, I did not add type casting where the default value is already a bool or where a `not` is present to perform the conversion.


pseudotensor (approved):
Thanks!

PR #337, opened 2023-06-25
Handle exceptions better when doing multi-view model lock, and don't block good endpoints

pseudotensor:
Handle exceptions better when doing multi-view model lock, so don't hang (quickly check if real connection there) and don't block or fail good endpoints in chat view.  Show exceptions for bad models in chat exceptions


pseudotensor:
image

image

PR #341, opened 2023-06-26
Queue input to avoid fresh submit using input_list at click/enter time, else truncates result because it uses input_list from time of click

pseudotensor:
- [x] Takes first enter/click, and gives chance while generating to take another "ask anything" entry, but once hit click/submit on that new one while generating, starts to queue.
- [x] Print out exception to logs for timeout streamer
- [x] Move chatbot output None -> '' earlier so done from start cleanly.
- [x] Clean-up (so faster) non- model_lock case and undo for both cases.

Otherwise, inputs (input_list that includes text_outputs that are the chatbot histories) at time of enter/click are used instead.

image

image

No truncation:

image


pseudotensor:
```
(alpaca) jon@gpu:/data/jon/h2o-llm$ pytest -s -v tests/test_client_calls.py &> clientcalls_log.txt

=========== 22 passed, 18 skipped, 2 warnings in 3104.63s (0:51:44) ============
(END)

```

Issue #282, opened 2023-06-13
E   ImportError: cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_' 


pseudotensor:
```
for fil in requirements.txt reqs_optional/requirements_optional_langchain.txt reqs_optional/requirements_optional_gpt4all.txt reqs_optional/requirements_optional_langchain.gpllike.txt ; do pip install -r $fil ; done
```
then for client:
```
cd client
make setup
make lint
make build
# install (choose version if multiple builds in dist directory)
pip install dist/h2ogpt_client-*-py3-none-any.whl
# test
cd ..
pytest -s -v --forked client
```

```
E   ImportError: cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_' (/home/jon/miniconda3/envs/h2ollm/lib/python3.10/site-packages/urllib3/util/ssl_.py)
```

Seems order matters, like installing again requirements.txt helps.  Can you figure out please?

Once pip install client wheel, basic `generate.py` calls fail:

```
(env) arno@rippa:/nfs4/llm/h2ogpt(main)$ python generate.py --base_model=eugenepentland/Minotaur-13b-Landmark --trust_remote_code=True
Traceback (most recent call last):
  File "/nfs4/llm/h2ogpt/generate.py", line 37, in <module>
    from accelerate import init_empty_weights, infer_auto_device_map
  File "/nfs4/llm/h2o-llm/env/lib/python3.10/site-packages/accelerate/__init__.py", line 3, in <module>
    from .accelerator import Accelerator
  File "/nfs4/llm/h2o-llm/env/lib/python3.10/site-packages/accelerate/accelerator.py", line 33, in <module>
    from .checkpointing import load_accelerator_state, load_custom_state, save_accelerator_state, save_custom_state
  File "/nfs4/llm/h2o-llm/env/lib/python3.10/site-packages/accelerate/checkpointing.py", line 24, in <module>
    from .utils import (
  File "/nfs4/llm/h2o-llm/env/lib/python3.10/site-packages/accelerate/utils/__init__.py", line 109, in <module>
    from .launch import (
  File "/nfs4/llm/h2o-llm/env/lib/python3.10/site-packages/accelerate/utils/launch.py", line 23, in <module>
    from ..commands.config.config_args import SageMakerConfig
[...]

PR #349, opened 2023-06-27
Client revamp

this:
### This PR,
- Unifies the text completion and chat completion APIs so that they both have a similar structure.
```py
# text completion API
text_completion = client.text_completion.create()
response = await text_completion.complete("Hello world")
```
```py
# chat Completion API
chat_completion = client.chat_completion.create()
response = await chat_completion.chat("Hey!")
print(response["user"])
print(response["gpt"])
```
- Adds `async` support to the chat completion API. Now the APIs are "`async` first".
- Reduce code duplication, especially parameter duplication between async and sync APIs.
- Improves tests.
  - Now you can run Client tests within the h2oGPT Python environment.
- Improves API docs.
- Improves README.
- Upgrades `gradio-client` to v0.2.7.
- Fixes https://github.com/h2oai/h2ogpt/issues/282

#### Tested against h2oGPT in latest `main` 15dda4a2f8f2631128e7f8edabd8cd894760bebd


client/Makefile
```diff
@@ -18,8 +18,18 @@ lint: generate_sources
 
 .PHONY: test
 test: generate_sources
-	poetry run pytest
+	poetry run pytest -r A
+
+.PHONY: test_with_h2ogpt
+test_with_h2ogpt: H2OGPT_START=python generate.py --base_model=h2oai/h2ogpt-oig-oasst1-512-6_9b --load_8bit=True
```
pseudotensor, line 24: It's best if a single `pytest -s -v tests` command can launch all tests.  I still don't understand why need separate env etc.  If only dependency for client itself is gradio client, then running as part of normal tests with h2oGPT env should be normal.

pseudotensor (approved):
Confirmed readme.md steps and tests pass.  Thank you!

PR #392, opened 2023-07-05
Add test to the Client validate parameters order with h2ogpt server

this:
TODO

- [ ] Refactor `generate.py` and move `eval_func_param_names` into a separate file.


pseudotensor (approved):
thanks!

PR #477, opened 2023-07-17
[DevOps] Update README for docker runtime image consumption

ChathurindaRanasinghe:


achraf-mer (approved):

achraf-mer:
Thanks @ChathurindaRanasinghe, let's also use simple semver docker version tags for every release.

PR #487, opened 2023-07-18
Minor Shell Script Changes

slycordinator:
Changed shell scripts for portability:
1) Changed to use /bin/sh instead of /bin/bash

Scripts all used posix-compliant commands, so /bin/sh will work and this way the scripts will run on a system that does not have bash

2) Changed backtick `` usage to "$()"
The former is largely deprecated.

3) Added quotes to variables as recommended by shellcheck utility


pseudotensor (approved):
thanks!

Issue #522, opened 2023-07-21
Create tests for long context


arnocandel:
goal is to test https://github.com/h2oai/h2ogpt/issues/360

- [ ] basic sanity test (check if can retrieve 1 secret at M-th token of N total tokens)
- [ ] implement more complicated tests, like https://github.com/DachengLi1/LongChat/tree/longeval/longeval

PR #524, opened 2023-07-21
Add long context tests and fix tokenizer truncation for activated rope_scaling

arnocandel:
Implements #522

### DIFF
--- a/src/gen.py
+++ b/src/gen.py
@@ -1041,7 +1041,7 @@ def get_model(
     if config is not None and tokenizer_loader is not None and not isinstance(tokenizer_loader, str):
         tokenizer = tokenizer_loader.from_pretrained(tokenizer_base_model, **tokenizer_kwargs)
         # sets raw (no cushion) limit
-        set_model_max_len(config, tokenizer, verbose=False)
+        set_model_max_len(config, tokenizer, verbose=False, rope_scaling=rope_scaling)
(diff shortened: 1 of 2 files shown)


arnocandel:
image

arnocandel:
currently fails with junk output
image
image

src/gen.py
```diff
@@ -1313,27 +1313,33 @@ def get_hf_model(load_8bit: bool = False,
...
+    rope_scaling_factor = 1
+    if rope_scaling:
+        rope_scaling_factor = rope_scaling.get('factor')
+        assert isinstance(rope_scaling_factor, int)
     if reward_type:
         # limit deberta, else uses too much memory and not worth response score
         tokenizer.model_max_length = 512
     if hasattr(config, 'max_seq_len') and isinstance(config.max_seq_len, int):
-        tokenizer.model_max_length = config.max_seq_len
+        tokenizer.model_max_length = config.max_seq_len * rope_scaling_factor
```
pseudotensor, line 1330: The max_seq_len should already be correct when loading the model, if not it should be done there instead of this factor passed through.

pseudotensor: I don't actually see any issue, it's already correct.

pseudotensor: https://github.com/h2oai/h2ogpt/commit/46720496234077e1e532992612b3f32174b87d23

pseudotensor (approved):

PR #543, opened 2023-07-26
Document how to disable chroma telemetry

mmalohlava:
The change enables Heap telemetry in Gradio UI to be able to improve
user experience.
The default tracking app id point to Heap project `Main` and its `Development` environment.

For any H2O.ai deployment we should use tracking app id from Heap
project `Main` and environment `Production`. @mmalohlava or from
@mtanco.

The analytics DO NOT track any prompts, only user interactions in UI
(click on buttons, links).

### DIFF
--- a/src/gradio_runner.py
+++ b/src/gradio_runner.py
@@ -52,8 +52,9 @@ fix_pydantic_duplicate_validators_error()
 
 from enums import DocumentSubset, no_model_str, no_lora_str, no_server_str, LangChainAction, LangChainMode, \
     DocumentChoice, langchain_modes_intrinsic
-from gradio_themes import H2oTheme, SoftTheme, get_h2o_title, get_simple_title, get_dark_js, spacing_xsm, radius_xsm, \
-    text_xsm
+from gradio_themes import H2oTheme, SoftTheme, get_h2o_title, get_simple_title,\
+    get_dark_js, get_heap_js, wrap_js_to_lambda,\
+    spacing_xsm, radius_xsm, text_xsm
 from prompter import prompt_type_to_model_name, prompt_types_strings, inv_prompt_type_to_model_lower, non_hf_types, \
     get_prompt
 from utils import flatten_list, zip_data, s3up, clear_torch_cache, get_torch_allocated, system_info_print, \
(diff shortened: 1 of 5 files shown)


mmalohlava:
@johnygomez @mtanco pls check

pseudotensor (commented):
Needs to be disabled by default and enabled via CLI option only.  Any keys should be removed from repo and passed on CLI.

pseudotensor (changes requested):
1

src/gradio_themes.py
```diff
@@ -221,11 +221,23 @@ def get_simple_title(title, description):
...
+    """
+
+def get_heap_js(heapAppId: str) -> str:
+    return ("""globalThis.window.heap=window.heap||[],heap.load=function(e,t){window.heap.appid=e,window.heap.config=t=t||{};var r=document.createElement("script");r.type="text/javascript",r.async=!0,r.src="https://cdn.heapanalytics.com/js/heap-"+e+".js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(r,a);for(var n=function(e){return function(){heap.push([e].concat(Array.prototype.slice.call(arguments,0)))}},p=["addEventProperties","addUserProperties","clearEventProperties","identify","resetIdentity","removeEventProperty","setEventProperties","track","unsetEventProperty"],o=0;o<p.length;o++)heap[p[o]]=n(p[o])};"""
+           f"""heap.load("{heapAppId}");""")
+
+def wrap_js_to_lambda(*args: str) -> str:
+    newline = "\n"
+    return f"""
+        () => {{
```
johnygomez, line 240: nit: Why it needs to have double brackets?

src/gradio_runner.py
```diff
@@ -2476,7 +2477,8 @@ def count_chat_tokens(model_state1, chat1, prompt_type1, prompt_dict1,
                        ,
                        queue=False, api_name='stop' if allow_api else None).then(clear_torch_cache, queue=False)
 
-        demo.load(None, None, None, _js=get_dark_js() if kwargs['dark'] else None)
+        app_js = wrap_js_to_lambda(get_dark_js(), get_heap_js(get_heap_app_id()))
+        demo.load(None, None, None, _js=app_js)
```
johnygomez, line 2481: QQ: Just to double-check, when exactly is this script called on the UI side? Is it when the page is loaded? Note that the heap script is downloading and using additional scripts, which could be a problem if called in some weird way or sandboxed. I assume however it's tested and it's working.

johnygomez (approved):
LGTM, thanks Michal, I have just a few points to double-check.

src/gradio_themes.py
mmalohlava: string interpolation requires it - f-string "{}" is treated as a reference to variable.

mmalohlava:
@pseudotensor 
> Needs to be disabled by default and enabled via CLI option only. Any keys should be removed from repo and passed on CLI.
the whole point is to collect telemetry to improve our understanding of end-users.

arnocandel (approved):

pseudotensor (approved):

PR #547, opened 2023-07-27
feat: expose 'root_path' from gradio

jcatana:
Adds additional param '--root_path' to from gradio Blocks.launch() arguments.

I've found this super simple change is a really handy feature and wanted to give it back in case others found it useful. It allows you to place the gradio interface behind a URL ingress subpath like https://domain.com/some-llm/

Thanks!


pseudotensor (approved):
Thanks!

PR #549, opened 2023-07-27
Add custom entrypoint args as env vars

ChathurindaRanasinghe:
- Get default values for arguments through environment variables such that arguments can be passed via environment variables in helm charts.


EshamAaqib (approved):
LGTM

pseudotensor (changes requested):
There must be better way to handle this.  It makes the --help from Fire ugly and  not useful and is hard to read.

achraf-mer (changes requested):
can keep it simple, e.g.: https://github.com/h2oai/h2ogpt/pull/556

PR #556, opened 2023-07-27
Load args from env vars as long as var starts with H2OGPT_

achraf-mer:


src/gen.py
```diff
@@ -2822,7 +2823,7 @@ def entrypoint_main():
 
     python generate.py --base_model=h2oai/h2ogpt-oig-oasst1-512-6_9b
     """
-    fire.Fire(main)
+    H2O_Fire(main)
```
achraf-mer, line 2826: @ChathurindaRanasinghe I only changed here, can you please review and also check other places where env vars are needed. Thanks

pseudotensor:
Cool thanks!

pseudotensor (commented):
Thanks.  I was thinking of something related a month ago.  I wanted to be able to set any parameter of main() via ENV.  Thanks for this, since I think it takes care of that too.

What about just H2OGPT_x .  Why need _PARAM?  Also, ok to allow all lower-case too like we do in DAI?  Thanks!

src/gen.py
```diff
@@ -42,13 +42,14 @@
...
 from typing import Union
 
-import fire
 import torch
 from transformers import GenerationConfig, AutoModel, TextIteratorStreamer
 
 from prompter import Prompter, inv_prompt_type_to_model_lower, non_hf_types, PromptType, get_prompt, generate_prompt
 from stopping import get_stopping
 
+from src.utils import H2O_Fire
```
pseudotensor, line 51: I think this should just be `from utils` to be consistent.

achraf-mer:
> Thanks. I was thinking of something related a month ago. I wanted to be able to set any parameter of main() via ENV. Thanks for this, since I think it takes care of that too.
> 
> What about just H2OGPT_x . Why need _PARAM? Also, ok to allow all lower-case too like we do in DAI? Thanks!

ok, done in https://github.com/h2oai/h2ogpt/pull/556/commits/4b907339a62c057870e99c8e828b63316f2d6e42

no prioritization for now (i.e. all caps taking precedence) can do if needed.

also, `Fire` is not very lenient, for example, this works:
`H2OGPT_LOAD_EXLLAMA=True` or `--load_exllama=True`

but with lower case `t` it doesn't:
`H2OGPT_LOAD_EXLLAMA=True` or `--load_exllama=True`

src/gen.py
achraf-mer: make sense, done.

pseudotensor:
> > Thanks. I was thinking of something related a month ago. I wanted to be able to set any parameter of main() via ENV. Thanks for this, since I think it takes care of that too.
> > What about just H2OGPT_x . Why need _PARAM? Also, ok to allow all lower-case too like we do in DAI? Thanks!
> 
> ok, done in [4b90733](https://github.com/h2oai/h2ogpt/commit/4b907339a62c057870e99c8e828b63316f2d6e42)
> 
> no prioritization for now (i.e. all caps taking precedence) can do if needed.
> 
> also, `Fire` is not very lenient, for example, this works: `H2OGPT_LOAD_EXLLAMA=True` or `--load_exllama=True`
> 
> but with lower case `t` it doesn't: `H2OGPT_LOAD_EXLLAMA=True` or `--load_exllama=True`

Ok, I think you mean 2nd case to be "true".  ya, it's not toml, it's ok.

Ya, and no need for prioritization like in DAI IMO.  Thanks!

achraf-mer:
> > > Thanks. I was thinking of something related a month ago. I wanted to be able to set any parameter of main() via ENV. Thanks for this, since I think it takes care of that too.
> > > What about just H2OGPT_x . Why need _PARAM? Also, ok to allow all lower-case too like we do in DAI? Thanks!
> > 
> > 
> > ok, done in [4b90733](https://github.com/h2oai/h2ogpt/commit/4b907339a62c057870e99c8e828b63316f2d6e42)
> > no prioritization for now (i.e. all caps taking precedence) can do if needed.
> > also, `Fire` is not very lenient, for example, this works: `H2OGPT_LOAD_EXLLAMA=True` or `--load_exllama=True`
> > but with lower case `t` it doesn't: `H2OGPT_LOAD_EXLLAMA=True` or `--load_exllama=True`
> 
> Ok, I think you mean 2nd case to be "true". ya, it's not toml, it's ok.
> 
> Ya, and no need for prioritization like in DAI IMO. Thanks!

oops, that's right, sorry, updated.

pseudotensor:
Can you add a note and example here?

https://github.com/h2oai/h2ogpt/blob/main/docs/FAQ.md#what-envs-can-i-pass-to-control-h2ogpt

Thanks!

achraf-mer:
> 


> Can you add a note and example here?
> 
> https://github.com/h2oai/h2ogpt/blob/main/docs/FAQ.md#what-envs-can-i-pass-to-control-h2ogpt
> 
> Thanks!
> 
> We can keep all those older envs for now, don't want to break things if others using them.
> 
> But good to know how to generally set the env for any main() parameters.

added here: https://github.com/h2oai/h2ogpt/blob/am/load_args_from_env/docs/FAQ.md#what-envs-can-i-pass-to-control-h2ogpt

pseudotensor (approved):

src/gen.py
achraf-mer: please review post merge.

PR #579, opened 2023-07-30
Improve READMEs

mihado:
I struggled to get LLaMa2 working in GPU mode with a 1080Ti on Ubuntu 22.04. I think by tweaking the sequences of the READMEs people would discover the important instructions more easily.

* Highlight that LLaMa2 runs in CPU mode by default
* Improve links between README & docs/README

If this makes sense, I'll update the docs for Windows & Mac as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,24 +20,28 @@ Query and summarize your documents or just chat with local private GPT LLMs usin
 
 ### Getting Started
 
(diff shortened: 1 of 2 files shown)


pseudotensor:
Thanks, I incorporated some of these kinds of things in recent update.

Issue #586, opened 2023-07-31
Feature Request: Specify HTML Client


raymondtri:
I see that the unstructured file loader is there by default, it would be awesome if we could instead force sites to be evaluated with selenium (or playwright) instead of that only occurring if there is a problem with loading the url, especially when loading large javascript or pwa-type sites.

pseudotensor:
It already works by using those others as backup if the original failed to get any docs. But yes good to specify.

pseudotensor:
https://github.com/h2oai/h2ogpt/commit/091587f923ce3a35d154d8f2d2066750feee6af4

pseudotensor:
![image](https://github.com/h2oai/h2ogpt/assets/2249614/b40b4a61-d28a-408e-8901-50a0142cac6b)

PR #597, opened 2023-08-02
added in rudimentary specificity in terms of the url loader to use

raymondtri:
Let me preface this with the fact that python is not something I have much experience in at all. That being said, this PR addresses #586 and allows for overriding the unstructuredURLloader with selenium or playwright by means of environment variables (ONLY_SELENIUM / ONLY_PLAYWRIGHT). I figured environment variables would make this easier to implement in other areas/scripts, some kind of getter pattern may be even better but this works.

added in modifications to the get_docs_with_score to help increment down slightly faster instead of -10 each time (honestly not sure if this is working, but starting at 1000 and then decrementing 10 each time is not great if your store only has say 50 docs, it's a workaround to a workaround), as well as added in rudimentary specificity in terms of the url loader to use

### DIFF
--- a/src/gpt_langchain.py
+++ b/src/gpt_langchain.py
@@ -34,7 +34,7 @@ from gen import get_model, SEED
 from prompter import non_hf_types, PromptType, Prompter
 from utils import wrapped_partial, EThread, import_matplotlib, sanitize_filename, makedirs, get_url, flatten_list, \
     get_device, ProgressParallel, remove, hash_file, clear_torch_cache, NullContext, get_hf_server, FakeTokenizer, \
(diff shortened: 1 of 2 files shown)


src/gpt_langchain.py
```diff
@@ -2278,7 +2293,11 @@ def get_docs_with_score(query, k_db, filter_kwargs, db, db_type, verbose=False):
                     print("chroma bug: %s" % str(e), flush=True)
                 if k_db == 1:
                     raise
-                if k_db > 10:
+                if k_db > 500:
```
pseudotensor, line 2296: This does double iteration of k_db

pseudotensor:
Thanks!  I posted changes that account for this.

PR #618, opened 2023-08-05
Handle persistence of user states for personal/scratch spaces.

pseudotensor:
- [x] Where to put user information.  Can't use user_path to decide if shared or not, since can switch to no user path.
- [x] Have user folders
- [x] Allow guest even if login
- [x] How to run something at start, so reload chat state etc.? https://github.com/gradio-app/gradio/issues/5138
- [x] use get_source_files_given_langchain_mode like way to get update ui when not in dbs yet
- [x] Return from get_source_files_given_langchain_mode db state updates so not slow next time


pseudotensor:
image

PR #624, opened 2023-08-07
[DevOps] Add Packer script for GCP

ChathurindaRanasinghe:
- Add Packer script to machine images for GCP
- Add provisioning script to setup Python, CUDA, h2oGPT
- Added a h2oGPT as a service to run the llama model by default

Parent Issue: 
- https://github.com/h2oai/h2o-ops/issues/370


cloud/packer/gcp/install_h2ogpt.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash -e
+
+cd h2ogpt
```
achraf-mer, line 3: any way we can reuse this code and create a script that is common to both the Dockerfile and this GCP script?

ChathurindaRanasinghe: Yeah @achraf-mer

ChathurindaRanasinghe:
Moved to https://github.com/h2oai/h2ogpt/pull/788

PR #659, opened 2023-08-13
some fixes for docker run

zba:
I found it not works "out of the box" and have some mistakes in config, so I fixed some, please consider to check.


achraf-mer:
LGTM @zba thanks.

Issue #651, opened 2023-08-11
Docker Image Permission Issues


ffalkenberg:
Hello maintainers,

I've encountered a problem when trying to deploy the `h2ogpt` Docker image on a Kubernetes/OC cluster. The deployment fails due to a `Permission denied` error. I believe the root cause is the way the Docker image is set up.

**Issue Details:**

1. Python is installed and run from the `/root` directory. This is problematic as many Kubernetes environments restrict running containers as root for security reasons.
2. The Docker image does not seem to be set up in a manner that's conducive for professional deployments, especially in environments that have strict security policies.

**Deployment Configuration:**

We are using the following `deployment.yaml`:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: h2ogpt-deployment
  labels:
    app: h2ogpt
spec:
  replicas: 1
  selector:
    matchLabels:
      app: h2ogpt
  template:
    metadata:
      labels:
        app: h2ogpt
    spec:
      containers:
      - name: h2ogpt-container
        image: gcr.io/vorvan/h2oai/h2ogpt-runtime:latest
        ports:
        - containerPort: 7860
        env:
        - name: HOME
          value: /workspace/
        volumeMounts:
        - name: cache-volume
          mountPath: /workspace/.cache
        - name: save-volume
          mountPath: /workspace/save
        command:
          - tail
          - '-f'
          - /dev/null
      volumes:
      - name: cache-volume
        persistentVolumeClaim:
          claimName: cache-pvc
      - name: save-volume
        persistentVolumeClaim:
          claimName: save-pvc
```

**Temporary Solution:**

As a quick fix, we built an additional layer on top of the existing image:

```Dockerfile
FROM gcr.io/vorvan/h2oai/h2ogpt-runtime:latest

RUN chgrp -R 0 /root && \
    chmod -R g=u /root

RUN chgrp -R 0 /workspace && \
    chmod -R g=u /workspace
```

However, this is not a sustainable long-term solution.

**Request:**

I kindly request that the Docker image be restructured to avoid such permission issues, especially when deploying in environments like Kubernetes. Ideally, the image should not rely on the `/root` directory for operations and should be set up in a way that respects common security practices.

Thank you for your attention to this matter.

Dockerfile: [Link](https://github.com/h2oai/h2ogpt/blob/main/Dockerfile)
Dockerimage: [Link](https://console.cloud.google.com/gcr/images/vorvan/global/h2oai/h2ogpt-runtime)

[...]

PR #664, opened 2023-08-14
Fix docker permissions and allow using a non root user

achraf-mer:
Address https://github.com/h2oai/h2ogpt/issues/651


ChathurindaRanasinghe (approved):
LGTM Thanks Achraf

PR #666, opened 2023-08-14
the message copy button moved closer to top border and message padding increased to 16px

fazpu:
- the message copy button moved closer to top border
- message padding increased to 16px

image


pseudotensor (approved):
thanks!

PR #704, opened 2023-08-20
Fix connectivity bug for offline llama

Mins0o:
When attempting to run [TheBloke](https://huggingface.co/TheBloke)/[Llama-2-13B-chat-GGML](https://huggingface.co/TheBloke/Llama-2-13B-chat-GGML) offline, I was getting some connection related issues. After modifying these lines, it works. Tested on a few other models, and it seems like it works. Further testing may be needed.  

Offline Tested models
 - TheBloke/Llama-2-13B-chat-GGML
 - TheBloke/Llama-2-13b-Chat-GPTQ --> failed
   - requests.exceptions.ConnectionError: (MaxRetryError("HTTPSConnectionPool(host='huggingface.co', port=443): Max retries exceeded with url: /api/models/TheBloke
/Llama-2-13b-Chat-GPTQ/revision/
 - h2oai/h2ogpt-gm-oasst1-en-2048-falcon-7b-v3 (w/ and w/o scoring model)
 - h2oai/h2ogpt-gm-oasst1-multilang-2048-falcon-7b
 - TheBloke/Wizard-Vicuna-13B-Uncensored-GPTQ
 - TheBloke/WizardLM-13B-V1.0-Uncensored-GPTQ --> failed
   - requests.exceptions.ConnectionError: (MaxRetryError("HTTPSConnectionPool(host='huggingface.co', port=443): Max retries exceeded with url: /api/models/TheBloke/WizardLM-13B-V1.0-Uncensored-GPTQ/revision/

### DIFF
--- a/src/gen.py
+++ b/src/gen.py
@@ -1011,7 +1011,7 @@ def get_config(base_model,
             if raise_exception:
                 raise
             if 'not a local folder and is not a valid model identifier listed on' in str(
-                    e) or '404 Client Error' in str(e):
+                    e) or '404 Client Error' in str(e) or "couldn't connect" in str(e):
                 # e.g. llama, gpjt, etc.
                 # e.g. HF TGI but not model on HF or private etc.
                 if max_seq_len is None and base_model.lower() in non_hf_types:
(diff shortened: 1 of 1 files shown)


pseudotensor:
Thanks!  I tested offline with llama and made fixes similar to yours but additional fixes for URL handling.

Mins0o:
https://github.com/h2oai/h2ogpt/commit/671f33a0a5307f32bfbd1942166d5f8627cb8685  
Thank you! I'll test it.  
  
I'm enjoying h2ogpt quite a lot. Thank you for the project

Mins0o:
I was still getting the error inside the `FakeTokenizer()` when I freshly started my docker container.  
It seems like this is the culprit.  
https://github.com/h2oai/h2ogpt/blob/f64c4cd7049e83cdbcd8d1ca77cbafcd2e285f2d/src/utils.py#L1022-L1024  
And since  
 - a. it has to do with the `tiktoken` package  
 - b. the workaroud is fairly easy (run once online, then go offline)  

I think it is negligible at the moment.

pseudotensor:
I think there are issues with docker and location of files.  FYI @achraf-mer @ChathurindaRanasinghe @arnocandel 

It should not download things inside docker.  I think there are bugs with that.

pseudotensor:
If going fully offline, should be this way: https://github.com/h2oai/h2ogpt/blob/main/docs/README_offline.md

But if have docker and already had those or files from prior run, shouldn't be downloaded inside docker.

Mins0o:
Hmm... I see  
Maybe there is something about how I have set it up, as I didn't follow the instructions strictly and took a little bit of liberty of tweaking in the docker file and command.  
I would say the problem is solved for me.  
Thank you for the fast and throughly kind support!

pseudotensor:
Ya, it's ok to work-around by avoiding tiktoken if that works.  We use tiktoken to count tokens for GGML models to handle limits of the model, but if no tokenizer is present I think we skip all that and it might still work.

achraf-mer:
https://github.com/h2oai/h2ogpt/pull/715

PR #715, opened 2023-08-21
Improve offline caching

achraf-mer:
image

Because of https://github.com/openai/tiktoken/blob/5d970c1100d3210b42497203d6b5c1e30cfda6cb/tiktoken/load.py#L30-L35

so set the cache to under h2oGPT cache dir, and pre-load at least the default encoding to things work in offline mode.


pseudotensor (approved):
It's a small addition, so ok, but I guess in general good to have users map external full .cache folder that was previously loaded up.

PR #717, opened 2023-08-21
bugfix-dark

ffalkenberg:
passing dark as parameter didn't work, i assume the reason is the handling of the load_events

### DIFF
--- a/src/gradio_runner.py
+++ b/src/gradio_runner.py
@@ -3572,16 +3572,11 @@ def go_gradio(**kwargs):
         else:
             auth = None
             load_func, load_inputs, load_outputs = None, None, None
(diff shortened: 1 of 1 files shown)


mmalohlava:
@ffalkenberg thank you for the contribution!

I would recommend to check this PR: https://github.com/h2oai/h2ogpt/pull/696 - we started with similar fix to your, but @pseudotensor pointed out that there are some problematic paths. Based on my understanding of Gradio code, we have to shape JS (especially lambda function) in the way compatible with Gradio's Block's inputs. 

Please check - any feedback is welcomed!

pseudotensor:
Yes it should be fine now, closing.

PR #733, opened 2023-08-22
don't set envs, just keep the defaults from HOME env var

achraf-mer:


ChathurindaRanasinghe (approved):
LGTM

achraf-mer:
for reference 
image

Issue #669, opened 2023-08-14
Docker documentation


patrickhwood:
A few documentation additions would be useful:

1. How to use make_db.py with docker, e.g., it fails without --save_dir.
2. Need to also mount -v "${HOME}"/.cache:/root/.cache for the instructor-large model to be saved properly (or set TORCH_HOME in the Dockerfile).
3. docker runs the container as root and fails with --user=<my uid>
4. The database is stored in the local directory, so either db_dir_<db-name> should be mounted to /workspace or $PWD at the head of the git repo should be mounted *as* /workspace.  (This second option should be mentioned as a good way to debug the docker container and even perform development.)

achraf-mer:
Thanks for the feedback @patrickhwood, my comments inlined:

>1. How to use make_db.py with docker, e.g., it fails without --save_dir.

@pseudotensor any idea about above?

>2. Need to also mount -v "${HOME}"/.cache:/root/.cache for the instructor-large model to be saved properly (or set TORCH_HOME in the Dockerfile).

should be covered by https://github.com/h2oai/h2ogpt/pull/728.

>4. docker runs the container as root and fails with --user=

the default docker user is `h2ogpt` not root, can you please point out how you ran the container to get it in a failed state?

>5. The database is stored in the local directory, so either db_dir_ should be mounted to /workspace or $PWD at the head of the git repo should be mounted as /workspace. (This second option should be mentioned as a good way to debug the docker container and even perform development.)

we do recommend to mount this directory `-v "${HOME}"/save:/workspace/save` as per description here: https://github.com/h2oai/h2ogpt/blob/main/docs/README_DOCKER.md, this can be used for all the save operations to maintain files locally, @pseudotensor are you aware of any other files that we automatically save with no option to specify the target location? (meaning things that we could lose if we terminate the container)

pseudotensor:
@achraf-mer On this one:
```
How to use make_db.py with docker, e.g., it fails without --save_dir.
```
make_db doesn't use --save_dir.


However, I think the issue is being able to run not just generate.py but also src/make_db.py since that is a separate thing one can run.  Is that possible?  Can we document?

patrickhwood:
Sorry.  I mean --user_path.  And yes, the issue is running make_db.py with docker, since the main reason I'm using docker is to have a separate python environment for all h2ogpt operations, since I didn't bother with requirements.txt in order to run generate.py with docker.

This was more a request to expand the docker docs, than a problem report running the code.

[...]

PR #750, opened 2023-08-24
Fix make_db.py from docker and document in readme

achraf-mer:
Address https://github.com/h2oai/h2ogpt/issues/669#issuecomment-1688565748


pseudotensor (approved):
thanks!

Issue #762, opened 2023-08-26
First upload works but additional data upload attempts get stuck


ryanchesler:
Might be related to https://github.com/gradio-app/gradio/pull/5249

Seems like upload is only working on the first file. If I upload an image it will process correctly but then if I try to upload a second image it will just get stuck uploading. It does not seem like it is hitting any of the later python code, just stuck on the front end. Maybe need a submit button to actually trigger the action again or do something to reset the state of file.upload.

<img width="182" alt="Screen Shot 2023-08-26 at 4 08 08 PM" src="https://github.com/h2oai/h2ogpt/assets/26419230/1f9e5a10-4f8b-4f8e-a3c3-e529b0d8adcb">

pseudotensor:
I can't even upload anything.  I think things deadlocked, related to:

```
huggingface/tokenizers: The current process just got forked, after parallelism has already been used. Disabling parallelism to avoid deadlocks...
To disable this warning, you can either:
        - Avoid using `tokenizers` before the fork if possible
        - Explicitly set the environment variable TOKENIZERS_PARALLELISM=(true | false)
```


perhaps.  That's normal message, but maybe with other things it got deadlocked

pseudotensor:
I don't think it's the tokenizer thing above.  All other things like text upload work.

pseudotensor:
1ed7a113ebed791c2848f7bc2eff0d9ac48232f2

Idea before was to still allow user to see the url link or text block during upload, but it was just hiding it anyways, so removed them.

Every upload hangs after image upload, so will try this.

pseudotensor:
Yes, with your and my changes no longer hangs

tares003:
![image](https://github.com/h2oai/h2ogpt/assets/10930271/44303796-e0b5-440f-9fce-dcb421744ebe)
I still have issue eventhough i have disabled it, it just hangs.

pseudotensor:
Looks like hanging during parsing.  Might not be hanging, could hit slow thing like images etc.

You can do:

```
kill -s SIGUSR1 <pid>
```
for the busy <pid> and share stack trace from console.

PR #765, opened 2023-08-27
Rc/#762 fixes file upload hanging on UI

ryanchesler:
https://github.com/h2oai/h2ogpt/issues/762


ryanchesler:
Manually tested putting multiple things in file upload and url and paste and all seem to work correctly with this minor change

pseudotensor (approved):
let's see!  I think we could get rid of all make_interactive make_non_interactive.  They were more useful then there were buttons it would control.  Now I think progress on them handles things.

Issue #755, opened 2023-08-25
Commit 7b3966116ef8d2bcaf37e1504dc63ccaa2bec072 causes redownload of huggingface models


patrickhwood:
Changing ENV TRANSFORMERS_CACHE=/workspace/.cache/huggingface/hub/ to huggingface/transformers/ causes all models already in ~/.cache/huggingface/hub to be downloaded again into ~/.cache/huggingface/transformers.  This slows down startup and really shouldn't happen, as the existing cache is mounted from the docker command line like this in the examples:

`-v "${HOME}"/.cache:/workspace/.cache`

Workaround is to set TRANSFORMERS_CACHE back to /workspace/.cache/huggingface/hub/ on the docker command line with

`-e TRANSFORMERS_CACHE=/workspace/.cache/huggingface/hub/`

but shouldn't be necessary.

pseudotensor:
![image](https://github.com/h2oai/h2ogpt/assets/2249614/0b865ba7-b12a-471d-b9af-113d0ff2cd54)

achraf-mer:
we can do https://github.com/h2oai/h2ogpt/pull/733, this forces the envs to be the absolute defaults, although the current defaults adhere to https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables, https://huggingface.co/transformers/v4.0.1/installation.html#caching-models, and https://pytorch.org/docs/stable/hub.html#where-are-my-downloaded-models-saved, so maybe an inconsistency in HF (meaning setting to described defaults in doc is different than not setting env vars)

achraf-mer:
@patrickhwood as you mentioned, this migration issue can be mitigated by setting the cache dir to the old location with the `/hub/` subdir:
```
-e TRANSFORMERS_CACHE=/workspace/.cache/huggingface/hub/
```
however I didn't see the /hub/ dir mentioned in HF docs, to illustrate what i described in my previous comment, here is what I see in the HF docs, and what I believe to be the source of discrepancy (ref: from links in previous comment):

![image](https://github.com/h2oai/h2ogpt/assets/51244975/349a4e09-0c53-48bb-97f5-d78615b7d35b)
![image](https://github.com/h2oai/h2ogpt/assets/51244975/83e22ac2-d0ac-4850-9857-29ae41b8471e)
![image](https://github.com/h2oai/h2ogpt/assets/51244975/ace6ff12-499b-4a69-bc13-b741e64683e1)

patrickhwood:
[...]

PR #770, opened 2023-08-28
h2oGPT Helm Chart

EshamAaqib:
## Description 

Helm Chart for h2oGPT, which includes vLLM and TGI


pseudotensor (approved):
great!

ozahavi (approved):
lgtm

helm/h2ogpt-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,361 @@
...
+              port: http
+            {{- toYaml .Values.vllm.readinessProbe | nindent 12 }}
+          {{- end }}
+          resources:
+            {{- toYaml .Values.vllm.resources | nindent 12 }}
+          envFrom:
+            - configMapRef:
+                name: {{ include "h2ogpt.fullname" . }}-vllm-inference-config
+          env:
+          - name: TRANSFORMERS_CACHE
```
achraf-mer, line 338: `TRANSFORMERS_CACHE` not really needed, we noticed that sometimes when setting env vars, it ends up having side effects different than what we expect, so best to keep the defaults if not necessary to change.

See an example here: https://github.com/h2oai/h2ogpt/issues/755#issuecomment-1696734046

babytdream:
@EshamAaqib @pseudotensor @ozahavi @achraf-mer @arnocandel 
Hi, are you interested in this question about improving vllm and benefiting you in #794 ？

PR #780, opened 2023-08-30
doc(macos): pin `llama-cpp-python` version for support GGML

iam4x:
Was having:

```
gguf_init_from_file: invalid magic number 67676a74
error loading model: llama_model_loader: failed to load model from llama-2-7b-chat.ggmlv3.q4_0.bin

llama_load_model_from_file: failed to load model
```

- countless hours of running into issues, not understanding this magic number error
- this works, and we have metal support

image

### DIFF
--- a/docs/README_MACOS.md
+++ b/docs/README_MACOS.md
@@ -67,7 +67,7 @@ Supports CPU and MPS (Metal M1/M2).
 * Metal M1/M2 Only:  Install and setup GPU-specific dependencies to support LLaMa.cpp on GPU:
     ```bash
     pip uninstall llama-cpp-python -y
-    CMAKE_ARGS="-DLLAMA_METAL=on" FORCE_CMAKE=1 pip install -U llama-cpp-python --no-cache-dir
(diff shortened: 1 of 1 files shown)


pseudotensor (approved):
yes thanks.

slavag:
@pseudotensor Well, also got this, but I upgraded version of gpt4all to the latest and now it works with latest version of llama-cpp-python.

pseudotensor:
Yes, if you use latest llama-cpp-python and gpt4all it might break langchain integration, I didn't test.  It may work.  But then you can only do GGUF models

https://github.com/h2oai/h2ogpt/commit/70f995822b8462b14b9996d6f751a3cae720a9b3

slavag:
Tried with my documents Q&A - works ok

pseudotensor:
@slavag Thanks for report!

slavag:
@pseudotensor small correction with llama-cpp-python 0.1.81 GGML was working fine, but with 0.1.83 not working anymore and need conversion to gguf.

PR #781, opened 2023-08-30
Update README_DOCKER.md

arnocandel:


docs/README_DOCKER.md
```diff
@@ -223,7 +223,6 @@ docker run \
     --shm-size=2g \
     -p 7860:7860 \
     --rm --init \
-    --network host \
```
pseudotensor, line 226: This was required during testing in order to have vLLM talk to h2oGPT (i.e. one docker talk to another).  Else they couldn't talk to each other.

pseudotensor: It's possible there is wrongness still, but maybe @achraf-mer can help.

achraf-mer: if removed fro the client side (h2ogpt), then one need to specify the actual ip address instead of 0.0.0.0, so it can work, just would make having an example in docs hard as it would depend on some ip config.

pseudotensor: https://github.com/h2oai/h2ogpt/commit/731221eb

achraf-mer: server:
image

client:
image

pseudotensor: Ya, but if local IP is used inside image, then how does that reach out?  Then would have to use --network host I think, so it is on same network as host system, and each docker on different port

PR #787, opened 2023-08-31
DocTR handling of pdfs

ryanchesler:
added in an option to process pdfs with DocTR. Converts pdf pages to numpy arrays at 300 DPI and then passes to DocTR the same we do with images.

Returns each pdf as its own document with pages concatenated together. Unclear if I should treat each page as its own doc or something else to split the page information. 

Seems like some of the logic in the pdf handling needs a bit of reworking. Some of the conditionals were set up so if the previous attempts yielded no contents it would try to do processing of methods that werent turned on.


ryanchesler:
@pseudotensor Should we set things up so if the method is turned on in the expert setting the method is always run? or only run if previous methods in the hierarchy fail? Maybe need to do multiple options like with enable_pdf_ocr where people can select on, auto or off to determine if something is always run or only when other methods fail.

ryanchesler:
Also, im noticing for some reason the pdf doctr is ending up faster than the image doctr but a decent margin. 8 seconds in pdf vs 30 seconds on image. Not sure if this is related to the loading and unloading of the model. I might have to look into it a little more.

ryanchesler:
PDF doctr handling is in a good state now. I tried to reduce a bit of the redundant code. I noticed blip and doctr werent being put back on gpu after it was unloaded so I fixed that as well

src/gpt_langchain.py
```diff
@@ -1703,29 +1705,18 @@ def file_to_doc(file, base_path=None, verbose=False, fail_any_exception=False,
...
             if doctr_loader is not None and not isinstance(doctr_loader, (str, bool)):
-                doctr_loader.set_image_paths([file])
-                docs1c = doctr_loader.load()
-                docs1c = [x for x in docs1c if x.page_content]
-                add_meta(docs1c, file, headsize, parser='doctr_loader')
+                doctr_loader.load_model()
             else:
                 from image_doctr import H2OOCRLoader
                 doctr_loader = H2OOCRLoader()
-                doctr_loader.set_image_paths([file])
```
pseudotensor, line 1713: Oh awesome, was going to start looking and noticed you did all this!

src/gpt_langchain.py
```diff
@@ -1834,7 +1824,7 @@ def file_to_doc(file, base_path=None, verbose=False, fail_any_exception=False,
             doc1a = clean_doc(doc1a)
             add_parser(doc1a, 'PyMuPDFLoader')
             doc1.extend(doc1a)
-        if len(doc1) == 0 or use_unstructured_pdf:
+        if (len(doc1) == 0 and use_unstructured_pdf) or use_unstructured_pdf:
```
pseudotensor, line 1827: There's 2 things going on with these conditions.  1) Always use things as backup if no docs found.  2) "Force" readers if appropriate, as we discussed.  "use_unstructured_pdf" really means "force additional parsing using unstructured"

src/gpt_langchain.py
```diff
@@ -1848,7 +1838,7 @@ def file_to_doc(file, base_path=None, verbose=False, fail_any_exception=False,
             add_parser(doc1a, 'UnstructuredPDFLoader')
             # seems to not need cleaning in most cases
             doc1.extend(doc1a)
-        if len(doc1) == 0 or use_pypdf:
+        if (len(doc1) == 0 or use_pypdf) and use_pypdf:
```
pseudotensor, line 1841: Same here, pypdf is backup in case no docs found, always done, while use_pypdf really means force also to use.

But things should be (in immediately future sense) setup for "max_quality" to avoid pypdf if doing pymupdf etc. since redundant in most cases.

src/gpt_langchain.py
ryanchesler: reverted these. I understand now. We have a default path but then the user can force them on if they want. Its not a straight on off switch. If they are off then the default path is still used.

src/gpt_langchain.py
ryanchesler: reverted

ryanchesler:
Made those changes so they could be forced on and made the same change for doctr. I originally changed it because I was struggling to get the doctr to trigger because it would always hit the prior pdf handling steps even when I had them off and I thought it was wrong. Now I get it and updated it so any of them could be forced on.

src/gpt_langchain.py
ryanchesler: Would be good if we had some test set to see what the impact is when we dump in redundant processing of the same sources. Unclear if the cost is just runtime or if passing in both an OCR representation and a PDF handling of the same doc actually causes worse retrieval

pseudotensor:
When running the test `test_png_add` I hit things like this:
```
  File "/data/jon/h2o-llm/src/gpt_langchain.py", line 1734, in file_to_doc
    caption_loader.to(caption_loader.device)
AttributeError: 'H2OImageCaptionLoader' object has no attribute 'to'

```

ryanchesler:
Should be fixed now. I ran the tests and it passed more of them but it was still failing on some of them. Tried on main and it failed the same tests though. `AttributeError: 'NoneType' object has no attribute 'similarity_search'` and assertion errors for the driverslicense.jpeg

pseudotensor:
Ok, see this in tests:
```
  return fn(*args, **kwargs)
  File "/home/jon/miniconda3/envs/alpaca/lib/python3.10/site-packages/transformers/modeling_utils.py", line 2009, in to
    raise ValueError(
ValueError: `.to` is not supported for `4-bit` or `8-bit` bitsandbytes models. Please use the model as it is, since the model has already been set to the correct devices and casted to the correct `dtype`.
```

I don't see the db one.  But I can take a look at this one.

pseudotensor:
image

pseudotensor:
Running full suite now.

pseudotensor:
```
=========================== short test summary info ============================
FAILED tests/test_client_calls.py::test_client_chat_nostream_llama7b - Assert...
FAILED tests/test_client_calls.py::test_client_chat_stream_langchain_steps[100-256]
FAILED tests/test_client_calls.py::test_client_chat_stream_langchain_steps[100-2048]
FAILED tests/test_client_calls.py::test_client_summarization_from_url[-1-https://github.com/h2oai/h2ogpt]
FAILED tests/test_client_calls.py::test_fastsys[True-None-instruct_vicuna] - ...
FAILED tests/test_client_calls.py::test_fastsys[True-None-one_shot] - Asserti...
FAILED tests/test_client_calls.py::test_fastsys[False-None-instruct_vicuna]
FAILED tests/test_client_calls.py::test_fastsys[False-None-one_shot] - Assert...
FAILED tests/test_inference_servers.py::test_openai_inference_server[openai_azure_chat-False]
FAILED tests/test_inference_servers.py::test_openai_inference_server[openai_azure_chat-True]
FAILED tests/test_langchain_units.py::test_qa_wiki_db_chunk_hf_dbs_llama[chroma]
FAILED tests/test_langchain_units.py::test_make_add_db[weaviate-0] - Assertio...
FAILED tests/test_langchain_units.py::test_make_add_db[weaviate-1] - Assertio...
FAILED tests/test_langchain_units.py::test_url_more_subunit - assert 0 > 0
FAILED tests/test_long_context.py::test_gradio_long_context_uuid_key_value_retrieval[False-0.2-5000-{'type':'dynamic', 'factor':2}-h2oai/h2ogpt-4096-llama2-13b-chat]
FAILED tests/test_long_context.py::test_gradio_long_context_uuid_key_value_retrieval[False-0.2-6000-{'type':'dynamic', 'factor':2}-h2oai/h2ogpt-4096-llama2-13b-chat]
FAILED tests/test_manual_test.py::test_chat_context - NotImplementedError: MA...
FAILED tests/test_manual_test.py::test_upload_one_file - NotImplementedError:...
FAILED tests/test_manual_test.py::test_upload_multiple_file - NotImplementedE...
FAILED tests/test_manual_test.py::test_upload_url - NotImplementedError: MANU...
FAILED tests/test_manual_test.py::test_upload_arxiv - NotImplementedError: MA...
FAILED tests/test_manual_test.py::test_upload_pasted_text - NotImplementedErr...
FAILED tests/test_manual_test.py::test_no_db_dirs - NotImplementedError: MANU...
FAILED tests/test_manual_test.py::test_upload_unsupported_file - NotImplement...
FAILED tests/test_manual_test.py::test_upload_to_UserData_and_MyData - NotImp...
FAILED tests/test_manual_test.py::test_chat_control - NotImplementedError: MA...
FAILED tests/test_manual_test.py::test_subset_only - NotImplementedError: MAN...
FAILED tests/test_manual_test.py::test_add_new_doc - NotImplementedError: MAN...
FAILED tests/test_manual_test.py::test_model_lock - NotImplementedError: MANU...
FAILED tests/test_metrics.py::test_bleurt - ImportError: To be able to use ev...
FAILED tests/test_metrics.py::test_bertscore - ImportError: To be able to use...
FAILED tests/test_metrics.py::test_wiki_split - ImportError: To be able to us...
FAILED tests/test_requirements.py::test_requirements - ValueError: ['numpy==1...
= 33 failed, 1308 passed, 278 skipped, 6 xfailed, 1 xpassed, 35 warnings in 19326.23s (5:22:06) =
```

The manuals always fail.  Most of the rest are flaky tests or no packages for metric case.  requirements test fails in numpy due to installing other things.  test_openai_inference_server needs to use real account, WIP.

PR #819, opened 2023-09-08
enable_pdf_doctr in utils

ffalkenberg:
Fix: Utilize enable_pdf_doctr parameter in lg_to_gr function

Previously, the enable_pdf_doctr parameter was overlooked during server initialization. This commit ensures it's properly utilized within the lg_to_gr function.


pseudotensor (approved):
thanks!

PR #824, opened 2023-09-11
Bump Helm Version

EshamAaqib:
# Description

Bump Helm version to the latest build. 

CC - @ChathurindaRanasinghe We may have to automate this to bump the version, automatically on each build


achraf-mer (commented):
can we merge with changes from https://github.com/h2oai/h2ogpt/pull/825, using app version as default is incorrect for tgi. Thanks

PR #825, opened 2023-09-11
helm chart improvements 

achraf-mer:


helm/h2ogpt-chart/values.yaml
```diff
@@ -86,15 +94,29 @@ vllm:
     repository: gcr.io/vorvan/h2oai/h2ogpt-runtime
     pullPolicy: IfNotPresent
 
+  imagePullSecrets:
+    - name: docker-creds
```
EshamAaqib, line 98: I think would be better to leave this field empty

helm/h2ogpt-chart/values.yaml
```diff
@@ -86,15 +94,29 @@ vllm:
...
+
   storage:
     size: 512Gi
     class: ebs-csi
   
   overrideConfig:
+    CUDA_VISIBLE_DEVICES: "0,1"
```
EshamAaqib, line 105: Same here

helm/h2ogpt-chart/values.yaml
```diff
@@ -112,8 +134,15 @@ vllm:
     seccompProfile:
 
   resources:
+
   nodeSelector:
```
EshamAaqib, line 138: Same here

helm/h2ogpt-chart/values.yaml
```diff
@@ -112,8 +134,15 @@ vllm:
...
+    hac.h2o.ai/accelerator: training-gpu
+
   tolerations:
```
EshamAaqib, line 141: Same here

EshamAaqib:
IMO, overrideConfig, containerArgs, nodeSelector and tolerations should be left empty so it configurable during installation

achraf-mer:
> IMO, overrideConfig, containerArgs, nodeSelector and tolerations should be left empty so it configurable during installation

done, PTAL thanks.

pseudotensor (approved):

PR #827, opened 2023-09-11
landing screens components re-ordered on mobile screens

fazpu:
- landing screens components re-ordered on mobile screens
- the mobile screen width break point is consistent with observed gradio breakpoint - 463px
- the `col_container` element id was brought one level up so that both sidebar and tabs live inside it

### Mobile screens
image

image

### Wide screens
image


pseudotensor (approved):
Thanks!

PR #829, opened 2023-09-11
header styling changed on mobile screen

fazpu:
- header styling changed on mobile screen
- the qr code is hidden on mobile screen

### mobile screens
image

### wide screens
image


pseudotensor (approved):
thanks!

PR #830, opened 2023-09-11
prompt area and upload button adjusted for mobile screens

fazpu:
- prompt area and upload button adjusted for mobile screens
- upload button is smaller on mobile screens
- default prompt textarea padding increased so that the text does not flow under the upload button

### mobile - 0px - 463px
image

image

### wide screens
image


pseudotensor (approved):
Thanks!

PR #831, opened 2023-09-11
visible models don't have the remove-all button

fazpu:
- visible models:
  - are no longer in the accordion
  - don't have the remove-all button
  - contain "Select: " before the input part

image


pseudotensor (approved):
Works thanks!

PR #853, opened 2023-09-15
multiselection styling applied both to model and document selection

fazpu:
image

image

### DIFF
--- a/gradio_utils/css.py
+++ b/gradio_utils/css.py
@@ -152,4 +152,18 @@ def make_css_base() -> str:
         padding: 0 4px;
         margin-right: 2px;
     }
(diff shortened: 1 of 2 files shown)


pseudotensor:
I still see the pop-up:

image

fazpu:
@pseudotensor yes, I don't know how to disable the password popup on h2oGPT side. It should be doable on gradio side.

The PR just add the same multiselect styling to the document tab.
- there is the "Select" string before the input
- the tags are removed
- the remove all button is gone

pseudotensor:
31516022e12208465f42b55f65d17ad170cf3aab

PR #858, opened 2023-09-16
Web search and Agents

### LINKED ISSUE #699: Web search option support [SERP API]
Hi

Do we have an web search option support in h2oGPT offline model?
If not is there a way to have a SERP API support using langchain.
I am new to this domain so I am not sure if there is an existing option to do this or it needs to designed, but I want to be able to force GPT to lookup using SERP for real time queries.

pseudotensor:
- [x] Add SERPAPI but vastly improve compared to langchain code (needs SERPAPI_API_KEY and free or other paid plans for heavier use)  Fixes #699
- [x] Handle combination of search and normal docs well, with special prompting and include as part of docs list, but as highest priority since usually small but relevant if selected
- [x] update langchain
- [x] update gradio for filterable disable
- [x] Wait for gradio to fix dataframes in UI
- [x] Fix llama.cpp classes after langchain update, tests fail
- [x] Test langchain agents, note that langchain goes too strongly by exact matches for output parser based upon regexp etc., when should use embeddings or LLM as optional decider like nvidia guardrails etc.
- [x] Add search, collection, python, csv, pandas, json agents.  Currently only works well with OpenAI
- [x] Generalize stopping for local transformers based models based upon terminate_response items
- [x] Ensure langchain stop passed through to pipeline for transformer based models
- [x] Handle token limits when have combination of docs, context, chat history, iinput, instruction
- [x] avoid increasing top_k_docs for summarization (3->8) for public portal
- [x] Allow more smaller search items since usually small, 8 total seems too little, all docs will be lost
- [x] Add 256 reduced max_new_tokens to API
- [x] Prep for interpreter

image

image

image

image

image

image


pseudotensor:
Hallucinations for 13B:

image

PR #879, opened 2023-09-21
Add text_context_list to directly pass text lists to LLM to avoid db etc. steps if don't care about persisting state and just want LLM to use context as if uploaded docs

pseudotensor:


pseudotensor:
Relevant and several other tests pass

PR #901, opened 2023-09-27
Add additional volume mount to Helm

EshamAaqib:
Added `/workspace/.cache/huggingface/modules` to mounts


achraf-mer (approved):
change LGTM but shouldn't be needed if using the same security context.

EshamAaqib:
> change LGTM but shouldn't be needed if using the same security context.

Security context is the same, but fails with permission denied errors with the mount

achraf-mer:
Addressed in a different PR.

PR #902, opened 2023-09-27
Allow rw to /workspace data

achraf-mer:


pseudotensor (approved):
Goal would be to ensure docker build makes an image that doesn't care about prior runtime artifacts that would be later generated inside or outside docker (unless run in same location and then of course user of docker has to worry if mapped from outside those folders to inside and their permissions etc.)

Issue #864, opened 2023-09-18
#Feature request Cpu docker compose 


bolavefasfas:
Cpu docker compose file or instructions on how to run it on cpu in a docker contianer ?

pseudotensor:
@ChathurindaRanasinghe are the instructions for CPU much different?

ChathurindaRanasinghe:
@pseudotensor Yeah running on CPU is different since current docker-compose is for running on GPU with our Dockerfile based on CUDA base image.

zorgonred:
Following, need this for my M2 Mac

pseudotensor:
@achraf-mer @ChathurindaRanasinghe This is still needed in general for our other products.

PR #927, opened 2023-10-05
Add docker compose for vllm and when running on CPU mode

achraf-mer:
Address: https://github.com/h2oai/h2ogpt/issues/864

### DIFF
--- a/docker-compose-vllm.yml
+++ b/docker-compose-vllm.yml
@@ -39,17 +39,7 @@ services:
     networks:
       - h2ogpt
     entrypoint: /h2ogpt_conda/vllm_env/bin/python3.10
-    command:
-      - -m
-      - vllm.entrypoints.openai.api_server
-      - --port=5000
-      - --host=0.0.0.0
-      - --model=${H2OGPT_BASE_MODEL}
-      - --tokenizer=${VLLM_TOKENIZER}
-      - --tensor-parallel-size=2
-      - --seed=1234
-      - --trust-remote-code
-      - --download-dir=/workspace/.cache/huggingface/hub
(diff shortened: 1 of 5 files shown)


docker-compose-vllm.yml
```diff
@@ -0,0 +1,72 @@
...
+      - -m
+      - vllm.entrypoints.openai.api_server
+      - --port=5000
+      - --host=0.0.0.0
+      - --model=${H2OGPT_BASE_MODEL}
+      - --tokenizer=${VLLM_TOKENIZER}
+      - --tensor-parallel-size=2
+      - --seed=1234
+      - --trust-remote-code
+      - --download-dir=/workspace/.cache/huggingface/hub
```
achraf-mer, line 52: can possibly move these configs to .env, but I don't see much benefit since not really re-used elsewhere.

pseudotensor: You mean it is a hard-coded example, but should be good.  does vLLM support CPU?  I thought people asking for CPU were asking just for h2oGPT at least.

achraf-mer: still no vllm support for CPU afaik.
>You mean it is a hard-coded example,
yes, I could move all generate arguments config to .env file, I just didn't see the need to, since not reused elsewhere.

arnocandel: i will improve, main problem now is that pycharm base tests don't see the makefile envs, nor the .env since in wrong place. so only settings.py would be doable, but then confusing since .env/settings/makefile all have special powers.

pseudotensor (approved):
be good if jenkins built this too as nightly.  Is that the case?

achraf-mer:
> be good if jenkins built this too as nightly. Is that the case?

Current nightly does not use these compose files, but can.

Issue #536, opened 2023-07-24
imperfect streaming for summarize


pseudotensor:
* HF models go through `H2OTextIteratorStreamer` that has stop iteration sent at end of each block.  So after first block streamed, rest are not, and final answer is revealed at once

* non-HF models use `StreamingGradioCallbackHandler` that stops on `on_llm_end` and way langchain is written, the first reduction block only calls end at end of all uses, and then final block is a new llm call that is not streamed

![image](https://github.com/h2oai/h2ogpt/assets/2249614/bccff1dc-e9cd-4a05-a6f6-38b901770934)

Could avoid stop iteration and/or avoid join to always stream, but need new way to stop that's not llm end but overal task end.  If wrong, then will just sit in ETHread loop forever.

pseudotensor:
risky:

```
diff --git a/src/gen.py b/src/gen.py
index ccae7f4..796475f 100644
--- a/src/gen.py
+++ b/src/gen.py
@@ -2229,8 +2229,8 @@ class H2OTextIteratorStreamer(TextIteratorStreamer):
     def on_finalized_text(self, text: str, stream_end: bool = False):
         """Put the new text in the queue. If the stream is ending, also put a stop signal in the queue."""
         self.text_queue.put(text, timeout=self.timeout)
-        if stream_end:
-            self.text_queue.put(self.stop_signal, timeout=self.timeout)
+        #if stream_end:
+        #    self.text_queue.put(self.stop_signal, timeout=self.timeout)
 
     def __iter__(self):
         return self
diff --git a/src/gpt_langchain.py b/src/gpt_langchain.py
index 04f4deb..4f1ac22 100644
--- a/src/gpt_langchain.py
+++ b/src/gpt_langchain.py
@@ -2072,6 +2072,14 @@ def _run_qa_db(query=None,
         context_class_cast = NullContext if device == 'cpu' or have_lora_weights else torch.autocast
         with context_class_cast(device):
             if stream_output and streamer:
+                if isinstance(llm.callbacks, list) and len(llm.callbacks) > 0 and isinstance(llm.callbacks[0], StreamingGradioCallbackHandler):
+                    # combine_docs() in map_reduce.py does 2 blocks, each uses llm separately and each calls on_llm_end() once.
+                    num_stops_expected = 1 if langchain_action in [LangChainAction.QUERY.value] else 2
+                else:
+                    # FIXME: Stops every use of LLM, need to know ahead of time what summarization would do,
[...]

PR #935, opened 2023-10-09
Improve summarization and add extraction -- speed-up streaming

pseudotensor:
- [x] Give control over how much to fill context with max_input_tokens
- [x] Give control over how to handle documents, e.g. split and merge as required up to max_input_tokens / model_max_length
- [x] Give control over way documents are joined when doing split for split_and_merge.  Also controls way estimate tokens for documents joined by langchain
- [x] Fix use of token_max passed to load_summarize_chain()
- [x] Improve efficiency of streaming by avoiding redundant/repeats of yield of same text
- [x] Add or ensure parallel for vLLM at least when not doing streaming (async stuff)
- [x] Figure out why gradio->gradio gives full streaming of every map but rest don't, and still reduce doesn't appear until done
- [x] Add async parallel for gradio inference, since helium uses instead of direct call
- [x] Add document extraction (map only and just collect intermediate results)

Streaming issues that are not blocking for later:
 https://github.com/h2oai/h2ogpt/issues/536#issuecomment-1757376403
- [ ] Figure out why reduce does not stream
- [ ] Consider if possible to stream when doing async parallel, e.g. stream and collect into text blocks.  Or just show one stream but rest avoid stream.


client/h2ogpt_client/_core.py
```diff
@@ -120,6 +122,12 @@ def create(
         :param text_context_list: list of strings to use as context (up to allowed max_seq_len of model)
         :param docs_ordering_type: By default uses 'reverse_ucurve_sort' for optimal retrieval
         :param min_max_new_tokens: minimum value for max_new_tokens when auto-adjusting for content of prompt, docs, etc.
+        :param max_input_tokens: Max input tokens to place into model context for each LLM call
+                                 -1 means auto, fully fill context for query, and fill by original docuemnt chunk for summarization
```
arnocandel, line 126: docuemnt

pseudotensor: tx

pseudotensor:
image

PR #941, opened 2023-10-11
Hk/main/benchmark plots

hemenkapadia:
Interactive Plotly visualization of the llm GPU benchmarks showing the interaction between 
- ( backend x LLM (7b,13b,70b) x quantization x # of GPUS x GPUs with different Memory x Task (summary, generation)

We normalize all GPU performance with the global CPU baseline for the task ( as no significant difference in CPU performance across bits). So all GPU numbers are some multiple of the CPU performance baseline. Since we have a common baseline we can compare the boost the GPU or setup provides to task

Some interesting observations
- 8 bit quantization is worst in performance. 
   - If possible work on 16 bits (full) or if quantization is needed to fit in mem then use 4 bit
- RTX 4090 and RTX 6000 sometimes beat A100. Why recommend a costly GPU ?
- having more than 1 GPUS on a machine does not have a very significant impact on processing.


hemenkapadia:
image

pseudotensor (approved):
thanks!

PR #978, opened 2023-10-21
For codellama or other JSON friendly models, stack system prompt with instructions and give document chunks in json, and ask for json output.

pseudotensor:
- [x] Allow system prompt to refer to JSON template and docs to be in JSON format and output to be in JSON format following template, with query reminding to use JSON output, to generate JSON output with extra meta info.  Models cannot self-verify well, so not great, but still interesting playground.
- [x] Add other prompts to be used soon
- [x] Improve langchain prefix command so model with function tools doesn't use "python" tool and uses our specified tool, and it more often uses tool than thinking generally.

image


pseudotensor:
image


image

PR #982, opened 2023-10-21
Add gputil python package

tomkraljevic:
Add gputil to python requirements so vllm starts properly even when not running as root.


achraf-mer (approved):
LGTM.
Likely related to just openshit, as default user for our tests is a non root.

PR #998, opened 2023-10-25
relax max_new_tokens to be per prompt

pseudotensor:
For vllm/gradio inference only for now
Relax max_new_tokens to be per prompt
Instead of max of all prompts for summarization/extraction
To have more output for reduction, since been seeing reduction truncated despite not alot of input.

image

### DIFF
--- a/src/gen.py
+++ b/src/gen.py
@@ -3957,6 +3957,19 @@ def history_to_context(history, langchain_mode=None,
     return context1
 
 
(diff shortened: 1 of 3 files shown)


pseudotensor:
Ignore mistrals, but ok, but what is stopping things for some models?

image

pseudotensor:
Sometimes fine, same exact summarization  with top_k_docs=2 on same doc.  Unclear why happens, but probably requests or something hangs-up threading and stops other threads from generating.

image

PR #1010, opened 2023-10-27
Use migration-safe `/submit_nochat_api` for the ChatCompletion API

### LINKED ISSUE #995: make chat completion API migration safe
For chat completion the API has to match exactly between server and client, so if the server is some latest h2ogpt, the client has to match.  It's not migration safe.  Only the text completion API is migration safe.

I always update the client API when I change things, and it's fine.  You just have an old wrong version.
@this  can consider making the chat completion API migration safe.  i.e. there's no need to use chat=True for chat completion, the client could manage the conversation state as it already has to.

i.e. this: https://github.com/h2oai/h2ogpt/blob/79f9dd56fac11535c10e826fc25e75ca3d883034/client/h2ogpt_client/_core.py#L299 

can be set to False and the "chatbot" can be purely managed by client, instead of expecting back an update history of the chat from the server.  It's totally useless by design since the history anyways has to be updated by client before asking another question with that updated history.  Instead the client could just maintain the history, and then the nochat API can be used and be migration safe.

this:
Fixes #995

### DIFF
--- a/client/h2ogpt_client/_core.py
+++ b/client/h2ogpt_client/_core.py
@@ -1,7 +1,6 @@
 import ast
 import asyncio
-import typing
-from typing import Any, Dict, List, Optional, OrderedDict, Tuple, Union, ValuesView
+from typing import Any, Dict, List, Optional, OrderedDict, Union
 
 import gradio_client  # type: ignore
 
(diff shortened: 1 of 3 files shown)


pseudotensor (approved):
Cool thanks!  Did you run the tests or should I check?

this:
@pseudotensor I ran the tests. Also tested with https://llama.h2o.ai

PR #1024, opened 2023-10-31
Fix source file link

us8945:


docs/README_CLIENT.md
```diff
@@ -27,7 +27,7 @@ res = client.predict(str(dict(kwargs)), api_name='/submit_nochat_api')
 response = ast.literal_eval(res)['response']
 print(response)
 ```
-For other ways to use gradio client, see example [test code](../client_test.py) or other tests in our [tests](https://github.com/h2oai/h2ogpt/blob/main/tests/test_client_calls.py).  E.g. `test_client_chat_stream_langchain_steps3` etc. [tests](https://github.com/h2oai/h2ogpt/blob/main/tests/test_client_calls.py) use many different API calls for docs etc.
+For other ways to use gradio client, see example [test code](../scr/client_test.py) or other tests in our [tests](https://github.com/h2oai/h2ogpt/blob/main/tests/test_client_calls.py).  E.g. `test_client_chat_stream_langchain_steps3` etc. [tests](https://github.com/h2oai/h2ogpt/blob/main/tests/test_client_calls.py) use many different API calls for docs etc.
```
pseudotensor, line 30: Thanks.  I think you meant "src" not "scr"

us8945: Yes, I just pushed the fix.

pseudotensor (approved):
tx!

PR #1026, opened 2023-11-01
Refactor models API in the Client

this:
- Adds `_models.py` sub-module which contains `Models`, `Model` classes.
- Now available models can be listed as `client.models.list()` which is similar to the OpenAI API.
- Renames `visible_models` param in text/chat completion create method to `models` to be more like the OpenAI API https://platform.openai.com/docs/guides/gpt/chat-completions-api
- Now users can pass a `Model` object for the `models` param.


pseudotensor (approved):
thanks!

PR #1029, opened 2023-11-01
Rename `models` param to `model` in the Client

this:
- Renames the `models` param in text/chat completion create method to `model` because h2oGPT server doesn't let the Client choose multiple models.
- Improve text/chat completion tests to use the `model` parameter.

Tested with https://gpt.h2o.ai, https://llama.h2o.ai/


client/tests/test_client.py
```diff
@@ -80,17 +90,16 @@ def test_parameters_order(client, eval_func_param_names):
 
 
 @pytest.mark.parametrize("local_server", [True, False])
-def test_readme_example(local_server):
-    # self-contained example used for readme, to be copied to client/README.md if changed, setting local_server = True at first
+def test_readme_example(local_server, h2ogpt_key):
```
pseudotensor, line 93: How does this work for h2ogpt_key?  It's not defined no?  Also, I want to keep the test self-contained, no client passed in etc.

pseudotensor (approved):
tx!

PR #1031, opened 2023-11-01
Update README.md

surenH2oai:
fixed H2OGPT_H2OGPT_KEY to H2OGPT_KEY

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -50,7 +50,7 @@ def test_readme_example(local_server):
     if local_server:
         client = Client("http://0.0.0.0:7860")
     else:
-        h2ogpt_key = os.getenv('H2OGPT_H2OGPT_KEY')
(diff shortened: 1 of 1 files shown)


client/README.md
```diff
@@ -50,7 +50,7 @@ def test_readme_example(local_server):
     if local_server:
         client = Client("http://0.0.0.0:7860")
     else:
-        h2ogpt_key = os.getenv('H2OGPT_H2OGPT_KEY')
+        h2ogpt_key = os.getenv('H2OGPT_KEY')
```
this, line 53: This example should match the relevant test case in https://github.com/h2oai/h2ogpt/blob/11984e362438ac8dc08721103bd9f04d0e1f3511/client/tests/test_client.py#L93-L126
```suggestion
        h2ogpt_key = os.getenv("H2OGPT_KEY") or os.getenv("H2OGPT_H2OGPT_KEY")
```

PR #1034, opened 2023-11-01
Make Helm work with external and local LLM's

EshamAaqib:
# Description

Make Helm work with external and local LLM's. When both are enabled we should not add `inference_server` to env vars instead use `MODEL_LOCK`. This PR fixes that


ozahavi (approved):
LGTM

pseudotensor (approved):

achraf-mer (approved):

PR #1046, opened 2023-11-03
[Client] Add streaming support for the text completion API

this:
```py
client = Client("https://llama.h2o.ai", h2ogpt_key=h2ogpt_key)
text_completion = client.text_completion.create()
```
### Async
```py
async def test_async():
    response = await text_completion.complete(
        prompt="Write a poem about the Amazon rainforest. End it with an emoji.",
        enable_streaming=True,
    )
    async for token in response:
        assert token
        print(token, end="")
```
### Sync
```py
def test_sync():
    response = text_completion.complete_sync(
        prompt="Write a poem about the Amazon rainforest. End it with an emoji.",
        enable_streaming=True,
    )
    for token in response:
        assert token
        print(token, end="")
```

### DIFF
--- a/client/h2ogpt_client/_completion.py
+++ b/client/h2ogpt_client/_completion.py
@@ -2,7 +2,7 @@ import ast
 import collections
 from typing import Any, Dict, List, Optional, OrderedDict, Union
 
-from h2ogpt_client import _core
+from h2ogpt_client._gradio_client import GradioClientWrapper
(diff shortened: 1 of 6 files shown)


pseudotensor (approved):
Cool thanks!  Please add something in the readme for this.

pseudotensor:
@this You'll add to readme?

this:
Yep, will create a separate PR

PR #1053, opened 2023-11-05
Fixing syntax error in sed README_MACOS.md

amizzo87:


pseudotensor:
Thanks!  I decided to make an openvllm wheel instead.

PR #1057, opened 2023-11-06
[DevOps] Cloud Image Fixes

ChathurindaRanasinghe:
- Fixes for cloud images
- Fixed Nginx issues, add service files as provisioning scripts


cloud/packer/setup_environment.sh
```diff
@@ -23,7 +23,6 @@ echo -e "Package: *\nPin: origin nginx.org\nPin: release o=nginx\nPin-Priority:
 sudo DEBIAN_FRONTEND=noninteractive apt -y update
```
achraf-mer, line 23: how does this differ than https://github.com/h2oai/h2ogpt/blob/devops/chathurinda/cloud-image-fixes/docker_build_script_ubuntu.sh ?

ChathurindaRanasinghe: It installs CUDA

achraf-mer (approved):

PR #1058, opened 2023-11-07
Fix prompting in langchain pandas csv agents, missing format_instructions and uses mrkl prompt even if make class on top, and no way to work around

pseudotensor:


pseudotensor:
But models are not good enough, need to be pandas fine-tuned etc.  Can't even do basic understanding of columns by names etc that even gpt-3.5 turbo can do, but that's through function tools.

image


image

PR #1073, opened 2023-11-09
Parse files inside tar.gz

### LINKED ISSUE #912: request: parse files inside tar.gz archives
That's the standard unix archive format, sometimes uses tgz extension but tar can be used with other compression formats for ex tar.xz

Mathanraj-Sharma:
Fix #912 

image


pseudotensor (commented):
thanks!  please add a test in test_langchain_unite

pseudotensor (approved):

PR #1120, opened 2023-11-22
[HELM] Remove default values from overrideConfig 

EshamAaqib:
The default values in `overrideConfig` is causing h2oGPT to break when using external LLMs with MODEL_LOCK. This PR removes the default values and add it as a comment


ozahavi (approved):
LGTM

achraf-mer (approved):

PR #1157, opened 2023-12-02
Update audio_langchain.py to fix Youtube on Windows

cherrerajobs:
Previous code did not submit a precise directory to the transcriber on Windows. Please verify this does not break *unix environments. I interpret the previous code as placing files into linux root folder `/tmp`, however on Windows the code places it in the current directory, creating a new `tmp` folder. When the transcriber tries to load the audio file it searches on the current drive's root `/tmp` folder (if it even exists) just like on *unix systems, but will never find the file there. This edit forces a full directory string to be used that points to the current directory's `tmp` folder.

Additionally, you should know that the current code base on windows places all files, temporary or not, including databases, in the directory h2ogpt is launched from, see:
image

### DIFF
--- a/src/audio_langchain.py
+++ b/src/audio_langchain.py
@@ -380,7 +380,8 @@ class H2OAudioCaptionLoader(ImageCaptionLoader):
 
         # https://librosa.org/doc/main/generated/librosa.load.html
         if from_youtube:
-            save_dir = "/tmp/" + "_" + str(uuid.uuid4())[:10]
+            current_path = os.getcwd().replace(os.sep, '/') # change backslashes to forward slashes on Windows
+            save_dir = current_path+"/tmp/" + "_" + str(uuid.uuid4())[:10]
             loader = GenericLoader(YoutubeAudioLoader(self.audio_paths, save_dir), self.model)
             return loader.load()
         else:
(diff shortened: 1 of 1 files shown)


src/audio_langchain.py
```diff
@@ -380,7 +380,10 @@ def load(self, from_youtube=False) -> List[Document]:
 
         # https://librosa.org/doc/main/generated/librosa.load.html
         if from_youtube:
-            save_dir = "/tmp/" + "_" + str(uuid.uuid4())[:10]
+            current_path = ""
```
pseudotensor, line 383: Thanks.  indeed using hard-coded /tmp is not good idea.  I'll see if can generalize.

pseudotensor: Can you try https://github.com/h2oai/h2ogpt/commit/ae299f7649b66e7589a6c292b97f0ba6d5f0b38d that's in main now?

I avoided the hard-coded /tmp and used normal python temp file.  Should work.

cherrerajobs: I can confirm it works! It properly places the audio file in the local temporary folder at C:\Users\[User]\AppData\Local\Temp\tmp[random_id]\[audio file].m4a and then transcribes it.

This should also allow for the content to be routinely cleaned by the OS. I don't know whether the app or OS manages that, but it's all in the correct place now.

PR #1199, opened 2023-12-13
Add the missing dependency `python-doctr`

Maverobot:
This PR fixes the error below when uploading PDFs in the web interface.

```
back (most recent call last):
  File "/Users/la/Dev/h2ogpt/src/gpt_langchain.py", line 3231, in path_to_doc1
    res = file_to_doc(file,
  File "/Users/la/Dev/h2ogpt/src/gpt_langchain.py", line 3034, in file_to_doc
    from image_doctr import H2OOCRLoader
  File "/Users/la/Dev/h2ogpt/src/image_doctr.py", line 16, in <module>
    from doctr.utils.common_types import AbstractFile
ModuleNotFoundError: No module named 'doctr.utils'
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -54,6 +54,7 @@ python-magic-bin==0.4.14; sys_platform == "win32"
 openpyxl==3.1.2
 lm_dataformat==0.0.20
 bioc==2.0
+python-doctr==0.7.0
 
 # falcon
 einops==0.6.1


pseudotensor:
doct is installed differently using custom package.  It's done here for linux:

https://github.com/h2oai/h2ogpt/blob/main/docs/linux_install.sh#L45

For windows the one-click installer becomes too large (>2GB) for it to handle.  But for windows one can install those things too.  I just haven't yet updated the instructions for windows.

pseudotensor:
Does DocTR actually work after your change?  In linux it requires quite a few other changes.

pseudotensor:
That's the issue I was expecting.  In the linux docs you'll see that installed using conda.  So probably doctr fails and a different package is used.

You can try doing what is in linux docs and run 

```
conda install weasyprint pygobject -c conda-forge -y
pip install onnxruntime==1.15.0
```

pseudotensor:
The architecture we use is based upon our own fine-tuned models, not the pure open-source DocTR.

So you can install ours instead.

```
pip install -r reqs_optional/requirements_optional_doctr.txt
```

I don't think there is any reason why it won't work on Mac.

pseudotensor:
if they work for you, then yes.

@Mathanraj-Sharma can you review docs/linux_install.sh and see which things will work for mac that are left out of normal install?

PR #1206, opened 2023-12-14
docs: faq: document auth.json file format

Blacksuan19:
this PR adds the expected format for `auth.json` file, I couldn't find where this is documented, came to this format after checking [gradio_runner.py](https://github.com/h2oai/h2ogpt/blob/main/src/gradio_runner.py) and testing on the docker image


pseudotensor (approved):
Thanks!  I'll merge and then update with some extra notes.

PR #1219, opened 2023-12-18
Use transformers version of attention sinks: https://github.com/huggingface/transformers/releases/tag/v4.36.0

pseudotensor:
Use transformers version of attention sinks: https://github.com/huggingface/transformers/releases/tag/v4.36.0

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -672,14 +672,13 @@ Exllama is supported using `load_exllama` bool, with additional control using `e
 
 Attention sinks is supported, like:
 ```bash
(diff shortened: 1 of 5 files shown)


pseudotensor:
```
python generate.py --base_model=HuggingFaceH4/zephyr-7b-beta --attention_sinks=True --max_new_tokens=10000 --max_max_new_tokens=10000 --max_seq_len=4096 --top_k_docs=-1
```

10,000 new tokens even though only 3829 input and used all input ~ 4096

image


But tends to repeat references etc.

Issue #1207, opened 2023-12-14
Hide Section header/selection if only 1 option is enabled, or list is empty


Blacksuan19:
For example, if only the Query action is enabled then hide the Actions section (header and selection), or if all agents are disabled then hide the agents header and dropdown.

also, the option to select which subset modes to enable would be great, we only use the relevant mode, subsequently the section should be hidden if only one option is enabled.

I'm passing these arguments to `generate.py` on the docker image. 
```bash
          --visible_langchain_actions="['Query']" \
          --visible_langchain_agents="[]" \
```

visual reference for how it looks now with above arguments.
![image](https://github.com/h2oai/h2ogpt/assets/10248473/e04b40b2-a08e-4226-b412-75d87860f3c1)

pseudotensor:
Good idea.

PR #1224, opened 2023-12-19
hide action selection if only one action is enabled

Blacksuan19:
check https://github.com/h2oai/h2ogpt/issues/1207


pseudotensor (approved):
yes I noticed and forgot to fix thanks!

PR #1239, opened 2023-12-23
Rotate image before OCR/DocTR - WIP

pseudotensor:
Merging as-is, not used yet.  Need logic to decide, e.g. have LLM decide which is best to use.


pseudotensor:
fastfood jpg_aligned
receipt jpg_rotated
rotate-ex2 png_rotated

PR #1241, opened 2023-12-26
Update README.md

zxankit24:


pseudotensor:
thanks, I'll keep original so no confusion as if there were changes to policy.

PR #1255, opened 2024-01-03
[HELM] Add vLLM check when running as stack

EshamAaqib:
This PR will make h2oGPT runtime wait until vLLM is up and running when running in stack mode

Issue - https://github.com/h2oai/public-cloud-infrastructure/issues/2561

### DIFF
--- a/helm/h2ogpt-chart/templates/deployment.yaml
+++ b/helm/h2ogpt-chart/templates/deployment.yaml
@@ -156,9 +156,15 @@ spec:
             {{- toYaml .Values.h2ogpt.securityContext | nindent 12 }}
           image: "{{ .Values.h2ogpt.image.repository }}:{{ .Values.h2ogpt.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.h2ogpt.image.pullPolicy }}
-          command: []
+          command: ["/bin/sh", "-c"]
           args:
-            - /workspace/generate.py
+          - |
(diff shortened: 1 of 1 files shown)


ozahavi (approved):
LGTM

helm/h2ogpt-chart/templates/deployment.yaml
```diff
@@ -156,9 +156,15 @@ spec:
...
-          command: []
+          command: ["/bin/sh", "-c"]
           args:
-            - /workspace/generate.py
+          - |
+            until wget -O- http://localhost:5000/v1/models >/dev/null 2>&1;
+            do
+              echo "Waiting for inference service to become ready...";
+              sleep 2;
+            done
```
achraf-mer, line 166: can do an explicit status code check, if you are sure won't be redirected:
```suggestion
            bash -c 'while [[ "$(curl --insecure -s -o /dev/null -w ''%{http_code}'' http://localhost:5000/v1/models)" != "200" ]]; do echo "Waiting for inference service to become ready... (2sec)" && sleep 2; done'
```

achraf-mer (approved):

helm/h2ogpt-chart/templates/deployment.yaml
EshamAaqib: Done - https://github.com/h2oai/h2ogpt/pull/1255/commits/c13d7c65b15520c6d7a9972a2209d7bc6d46eb1d. Tested as well and seems to be working

PR #1263, opened 2024-01-05
Remove HYDE accordion outputs if present before giving history to LLM, and remove chat=True/False for prompt generation, hold-over and led to bugs in prompting for gradio->gradio

pseudotensor:
Remove HYDE accordion outputs if present before giving history to LLM, and remove chat=True/False for prompt generation, hold-over and led to bugs in prompting for gradio->gradio


pseudotensor:
Fixed gradio->gradio for handling system prompt + chat conversation etc.

image

PR #1281, opened 2024-01-12
update langchain to use 0.1.0

achraf-mer:


achraf-mer:
See: https://h2oai.slack.com/archives/C05MUTAHKU5/p1705074165449009?thread_ts=1705073212.277659&cid=C05MUTAHKU5, current version is good as is.

PR #1288, opened 2024-01-13
prepare the docker runtime to use a separate vllm image

achraf-mer:


pseudotensor (commented):
I mentioned we can't generally go down, some models allow 0.2.4, but others do not like Mixtral.

tomkraljevic:
should we put the vllm into it's own image at this point?

(keeping one same helm chart is ok.)

in general, for on-prem it's more important for it to be stable than super-duper-latest.

pseudotensor:
Yes, that's what I initially recommended, a separate vllm 0.2.4 to support old models we need for now.  0.2.6 is required for newer models regardless of the memory issue.

achraf-mer:
changed the Helm implementation so vllm image can be used via:

```yaml
vllm:
  image:
    repository: vllm/vllm-openai
    tag: v0.2.6
    pullPolicy: IfNotPresent
```

docker_build_script_ubuntu.sh
```diff
@@ -73,28 +73,6 @@ for enc in model_encodings:
 print('Done!')
 "
 
-############################################################
```
pseudotensor, line 76: Just to be clear, I don't want to do this either.  Having them together is very useful.  I'm only suggesting we have a separate 0.2.4 for now because of specific model-hardware combinations.

docker_build_script_ubuntu.sh
```diff
@@ -73,28 +73,6 @@ for enc in model_encodings:
...
-# vllm server
-export VLLM_CACHE=/workspace/.vllm_cache
-cd /h2ogpt_conda
-python -m venv vllm_env --system-site-packages
-# gputil is for rayWorker in vllm to run as non-root
-# below required outside docker:
-# apt-get install libnccl2
-/h2ogpt_conda/vllm_env/bin/python -m pip install https://github.com/vllm-project/vllm/releases/download/v0.2.7/vllm-0.2.7+cu118-cp310-cp310-manylinux1_x86_64.whl
```
pseudotensor, line 84: The vllm docker image is a basic image that doesn't include sufficient packages for required models like Mixtral.  So one cannot remove these other dependencies.

tomkraljevic: the h2ogpt image is really bulky at over 30 GB.

having the smaller inferencing image is a good goal, because they don't share so much.
it would be nice to find a way for that to work.

pseudotensor: A separate vllm image is ok plan, just the way it's being done won't work.  We need to build our own image, but if we need 0.2.4 and latest, we need to build both.

docker_build_script_ubuntu.sh
achraf-mer: I re-worked the install scripts so one can upgrade vllm by itself while re-using some docker layers

pseudotensor:
We'll just use vLLM releases for stable models, rely upon h2oGPT's vllm for latest breaking changes.

dimm0:
Will you fix the command field in the helm chart? It's still "/h2ogpt_conda/vllm_env/bin/python3.10" which doesn't exist in latest images

achraf-mer:
> Will you fix the command field in the helm chart? It's still "/h2ogpt_conda/vllm_env/bin/python3.10" which doesn't exist in latest images

@dimm0 just to confirm that images under this registry: `gcr.io/vorvan/h2oai/h2ogpt-runtime` are all in one, meaning they include vllm under `/h2ogpt_conda/vllm_env/bin/python3.10`,
whether images under `gcr.io/vorvan/h2oai/h2oai-h2ogpt-runtime` use a separate vllm at `gcr.io/vorvan/h2oai/h2oai-h2ogpt-vllm`.
the default registry in helm will reflect the correct registry at: https://github.com/h2oai/h2ogpt/blob/main/helm/h2ogpt-chart/values.yaml#L13 or https://github.com/h2oai/h2ogpt/blob/h2ogpte/helm/h2ogpt-chart/values.yaml#L13

dimm0:
I don't see /h2ogpt_conda/vllm_env in gcr.io/vorvan/h2oai/h2ogpt-runtime

```
[root@gateway ~]# docker run -it --entrypoint="" gcr.io/vorvan/h2oai/h2ogpt-runtime:0.2.1-822 /bin/sh
h2ogpt@4bca256c3b1a:~$ ls /h2ogpt_conda/
LICENSE.txt                  cmake/                       condabin/                    include/                     sbin/                        ssl/
_conda                       compiler_compat/             envs/                        lib/                         share/                       x86_64-conda-linux-gnu/
bin/                         conda-meta/                  etc/                         man/                         shell/                       x86_64-conda_cos7-linux-gnu/
```

achraf-mer:
> I don't see /h2ogpt_conda/vllm_env in gcr.io/vorvan/h2oai/h2ogpt-runtime
> 
> ```
> [root@gateway ~]# docker run -it --entrypoint="" gcr.io/vorvan/h2oai/h2ogpt-runtime:0.2.1-822 /bin/sh
> h2ogpt@4bca256c3b1a:~$ ls /h2ogpt_conda/
> LICENSE.txt                  cmake/                       condabin/                    include/                     sbin/                        ssl/
> _conda                       compiler_compat/             envs/                        lib/                         share/                       x86_64-conda-linux-gnu/
> bin/                         conda-meta/                  etc/                         man/                         shell/                       x86_64-conda_cos7-linux-gnu/
> ```

sorry wrong assumption on my end, addressing here: https://github.com/h2oai/h2ogpt/pull/1763, thanks

PR #1308, opened 2024-01-19
Update README_MACOS.md

antoninadert:
I realized that I was getting these errors :
`Must install DocTR and LangChain installed if enabled DocTR, disabling`

The instruction in Linux made it stop. I tried the same line in Mac and it worked too


pseudotensor (approved):
thanks!

PR #1328, opened 2024-01-23
exposing imagePullSecret and tag in values.yaml

robinliubin:
`imagePullSecret` and custom `tag` are actually supported in deployment.yaml, for user's reference, just exposing them in values.yaml


EshamAaqib (approved):
LGTM

PR #1336, opened 2024-01-25
Update QR code.

arnocandel:


pseudotensor (approved):
will merge and check

odai1001:
Is it possible to solve the problem of why there is no response in Arabic on ubuntu?

odai1001:
> _لم يتم تقديم أي وصف._

Is it possible to solve the problem of why there is no response in Arabic on ubuntu?

PR #1337, opened 2024-01-25
h2ogpt support namespaceOverride

robinliubin:
this is needed for HAIC/genai Uber helm chart, if user need to decide which namespace h2ogpt will be running in.

### DIFF
--- a/helm/h2ogpt-chart/values.yaml
+++ b/helm/h2ogpt-chart/values.yaml
@@ -1,5 +1,6 @@
 nameOverride: ""
 fullnameOverride: ""
+namespaceOverride: 
 
 h2ogpt:
   enabled: true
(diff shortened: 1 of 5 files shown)


helm/h2ogpt-chart/values.yaml
```diff
@@ -1,5 +1,6 @@
 nameOverride: ""
 fullnameOverride: ""
+namespaceOverride:
```
ozahavi, line 3:
```suggestion
namespaceOverride: ""
```

ozahavi (approved):
LGTM

helm/h2ogpt-chart/values.yaml
robinliubin: Thanks, updated

PR #1433, opened 2024-02-26
API Enhancement

JohnnyFoulds:
This merge request adds the following minor changes to the user interface:

- `src/gradio_runner.py`: Page Header - The functionality already exists to configure the page title via the CLI, but unfortunately, the page header is to "h2oGPT". This change sets the page header to the same as the page title.

Additional enhancements were made to allow user authentications via the API to allow h2oGPT to be used as a managed RAG service.

- `openai_server/backend.py`: If a username, password, and langchain_mode are provided in the following format in the user parameter: "username:password:langchain_mode", it allows for seamless RAG integration via the OpenAI API.
- `src/gradio_runner.py`: Add the option to use `login` from the Gradio API.


pseudotensor:
@JohnnyFoulds Thanks for the PR!  You saw I incorporated the main idea into the other PR, but done differently in terms of allowing any parameter passed.

JohnnyFoulds:
> @JohnnyFoulds Thanks for the PR! You saw I incorporated the main idea into the other PR, but done differently in terms of allowing any parameter passed.

I saw, thanks, I like your way better.

PR #1444, opened 2024-03-01
Image change style

pseudotensor:


mati-gp:
I might be wrong but I think there's a bug in this line https://github.com/h2oai/h2ogpt/blob/d5c2b493b230b9b40ebd46760207e44dba1381b1/src/vision/utils_vision.py#L243

If `set(image_models).difference(valid_imagegen_models)) == 0` then it means that all models in `image_models` belong to `valid_imagegen_models`, which is OK and therefore it should not raise an exception.

It should be corrected to
`if enable_image and len(set(image_models).difference(valid_imagegen_models)) != 0:`

PR #1489, opened 2024-03-19
Linux full installation tweak

juerware:
Some tweaks for the file docs/linux_install_full.sh

### DIFF
--- a/docs/linux_install_full.sh
+++ b/docs/linux_install_full.sh
@@ -23,25 +23,19 @@ if ! command -v conda &> /dev/null; then
     echo "Conda not found, installing Miniconda."
     wget https://repo.anaconda.com/miniconda/Miniconda3-py310_23.1.0-1-Linux-x86_64.sh
     bash ./Miniconda3-py310_23.1.0-1-Linux-x86_64.sh -b -u
-    source ~/miniconda3/bin/activate
-    conda init bash
-    conda deactivate
 else
     echo "Conda is already installed."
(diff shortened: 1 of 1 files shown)


pseudotensor:
Hi, can you saw what issues you encountered and what this improves?  Thanks!

juerware:
Hi, of course I explain the improvements.

1) ```while [ $CONDA_SHLVL -gt 0 ]; do conda deactivate; done``` There can be a recursion in conda and this command reverts any conda session
2) ```if conda env list | grep -w 'h2ogpt'; then``` "w" for word is more accurate because "h2ogptBLABLABLA" can be a conda session as well and "h2ogpt" may not be a conda session therefore, there is no need to delete a session which does not exists.
3) ```if ! command -v nvcc &> /dev/null; then``` NVIDIA CUDA Compiler Driver NVCC is required but not informed about it till it fails when installation is running, in this line it is controlled.
4) ```if [ ! "$nvccVersion" == "12.1" ] && [ ! "$nvccVersion" == "11.8" ]; then``` h2ogpt documentation is based on these 2 versions of conda therefore, if these versions are not present in the system it advises about it and the user has the option whether to continue.
5) ```if [ "${nvccVersion}" == "11.8" ]; then``` variables in the system needs to take different values, standard variables are taken from cuda version 12.1

If you need more information, let me know and I explain myself as better as I can.

Thanks in advance.

juerware:
do you need anything else?

PR #1520, opened 2024-04-01
Add docai inference server capability

ryanchesler:
Adds query of docai server for available models and adds ability to pass model name to endpoint to tell endpoint which prompt embedding to use.


ryanchesler:
Makes it possible to do inference with https://github.com/h2oai/noria/blob/71f9bb998df90c527475a47fefeb945e322b63e1/src/inference_engine.py

src/evaluate_params.py
```diff
@@ -83,6 +83,7 @@
 
                          'image_file',
                          'image_control',
+                         'base_model'
```
pseudotensor, line 86: Hi, we use `visible_models` for this.  Even though it sounds plural, for API it can just be single string.

PR #1548, opened 2024-04-10
Do not force hf_transfer when explicitly set to false

achraf-mer:
Port https://github.com/h2oai/h2ogpt/pull/1547 to `main`.

### DIFF
--- a/src/gen.py
+++ b/src/gen.py
@@ -33,7 +33,7 @@ try:
 except (PackageNotFoundError, AssertionError):
     have_hf_transfer = False
 
-if have_hf_transfer:
+if have_hf_transfer and not os.getenv('HF_HUB_ENABLE_HF_TRANSFER', 'None') == '0':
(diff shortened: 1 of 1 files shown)


src/gen.py
```diff
@@ -33,7 +33,7 @@
 except (PackageNotFoundError, AssertionError):
     have_hf_transfer = False
 
-if have_hf_transfer:
+if have_hf_transfer and not os.getenv('HF_HUB_ENABLE_HF_TRANSFER', 'None') == '0':
```
pseudotensor, line 36: This will lead to it not being set if wasn't set.  That's opposite of what I want, doesn't make sense to me.

achraf-mer: 1. if not set, then condition equivalent to `and True:`
2. is set to 1 then condition will be equivalent to `and True:`
3. if set to 0 then condition will be equivalent to `and False:`

the code change is only to make sure `HF_HUB_ENABLE_HF_TRANSFER` is not set to 1 if we detect it as `0`.

PR #1610, opened 2024-05-10
[HELM] Fixes - Add Args when running h2oGPT only

EshamAaqib:


achraf-mer (approved):

EshamAaqib:
This is needed when running only h2oGPT without vLLM or TGI in cluster, Especially when depending completely on external LLM endpoints. When only h2oGPT is enabled it would fail to run since `args` are missing. This PR fixes this

ozahavi (approved):
LGTM

Issue #1616, opened 2024-05-14
umbrella podSecurityContext null values are always overwritten by sub-chart default values


robinliubin:
when install on openshift using umbrella chart.
umbrella can NOT nullify podSecurityContext, if set to null, it will be override by h2oGPT default values.

details in: https://github.com/h2oai/platform-charts/issues/260

need to set below values to empty, not default 1000:
https://github.com/h2oai/h2ogpt/blob/6824d21afc8633dc678e0f4dcdd0e450827faf91/helm/h2ogpt-chart/values.yaml#L69-L73
to 
```yaml
  podSecurityContext:
    runAsNonRoot: true
    runAsUser:
    runAsGroup:
    fsGroup:

```

PR #1618, opened 2024-05-14
set podSecurityContext to null, so umbrella can overwrite on openshift

robinliubin:
when deployed on openshift these values need to be nullified.

https://github.com/h2oai/h2ogpt/issues/1616
https://github.com/h2oai/platform-charts/pull/261


EshamAaqib:
Since default values are removed we will need to add them here as well https://github.com/h2oai/h2ogpte/blob/main/skaffold.yaml

pseudotensor (approved):

PR #1677, opened 2024-06-10
Reuse docs/linux_install.sh for h2ogpte image

achraf-mer:


linux_install_wolfi.sh
```diff
@@ -161,3 +161,6 @@ print('Done!')
 
 # mitigate CVE-2024-22423
 pip install yt-dlp==2024.4.9 --no-deps --no-cache-dir
+
+# mitigate CVE-2024-22195, CVE-2024-34064
+pip install jinja2==3.1.4 --no-cache-dir -c reqs_optional/reqs_constraints.txt
```
pseudotensor, line 166: Locally this is already the version I have, so must be already fixed I'd guess.

linux_install_wolfi.sh
```diff
@@ -161,3 +161,6 @@ print('Done!')
 
 # mitigate CVE-2024-22423
 pip install yt-dlp==2024.4.9 --no-deps --no-cache-dir
```
pseudotensor, line 163: I see this added here: https://github.com/h2oai/h2ogpt/commit/219fc57e88f44afdbc45a91d5a78f90c7b4d2c20  but what is the CVE?  This seems like a very old version, are you saying latest is broken?

pseudotensor: Or change https://github.com/h2oai/h2ogpt/blob/h2ogpte/linux_install_wolfi.sh#L36 instead to latest version for now until I change things.

linux_install_wolfi.sh
achraf-mer: no it is still at old version, but can remove and rebuild to see.

linux_install_wolfi.sh
achraf-mer: I removed and added requirements_optional_image.txt

pseudotensor (approved):
Thanks, seems good start.  I planned on looking too, but thanks for doing already this part.

PR #1733, opened 2024-07-11
Update nvidia-container-toolkit link

MagicalKyaru:
Update the nvidia container toolkit link according to the updated nvidia documentation. It allows the installation method to work properly on more diverse distributions, such as Ubuntu 24.04.

https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html


pseudotensor (approved):
thanks!

PR #1759, opened 2024-07-22
Update helm with other entries that are useful or required for h2oGPTe

pseudotensor:


helm/h2ogpt-chart/values.yaml
```diff
@@ -66,7 +66,37 @@ h2ogpt:
   #   max_new_tokens: 1024
 
   overrideConfig:
-
+    visible_login_tab        : False
```
EshamAaqib, line 69: Might have to remove the extra spacing ex visible_login_tab: False

EshamAaqib (approved):
LGTM apart from the comment

PR #1761, opened 2024-07-22
Port the internvl helm changes to the 'main' branch

achraf-mer:
Backport https://github.com/h2oai/h2ogpt/pull/1694/files to `main`.

### DIFF
--- a/helm/h2ogpt-chart/values.yaml
+++ b/helm/h2ogpt-chart/values.yaml
@@ -242,5 +242,49 @@ vllm:
   podLabels: {}
   autoscaling: {}
 
+internvl:
+  enabled: false
+  replicaCount: 1
+
+  image:
+    repository: gcr.io/vorvan/h2oai/h2oai-h2ogpt-internvl
+    tag:
+    pullPolicy: IfNotPresent
+
+  podAffinity:
+    # -- Set hostname and zone to true for pod affinity rules based on hostname and zone.
+    # hostname:
+    # zone:
+
+  storage:
+    size: 512Gi
+    class:
+    useEphemeral: true
+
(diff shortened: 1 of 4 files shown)


pseudotensor (approved):
tx!

helm/h2ogpt-chart/values.yaml
```diff
@@ -242,5 +242,49 @@ vllm:
   podLabels: {}
   autoscaling: {}
 
+internvl:
```
ozahavi, line 245: @achraf-mer why are we calling this `internvl` and not `lmdeploy`?

achraf-mer: no technical reason, I just used the name from the dockerfile thinking it would line up with the image we use. But if you or @pseudotensor think a different name would be more descriptive I can change.

ozahavi: I think it should be `lmdeploy` just like we have `tgi` and `vllm`.

pseudotensor: It's better as lmdeploy.  I just stuck on internvl because was model name I initially tried and lmdeploy are tightly related to the internvl team.

achraf-mer: ok, let's do it as part of this then: https://github.com/h2oai/h2ogpt/pull/1764, PTAL, thanks.

achraf-mer: the image repo is still called internvl but can change that too if needed.

ozahavi: That would be nice for the consistency

Issue #1774, opened 2024-07-30
intfloat_multilingual-e5-large-instruct and file sentence_xlnet_config.json


juerware:
I have executed the following command line:
```shell
python generate.py \
    --base_model=mistralai/Mistral-Nemo-Instruct-2407 --use_gpu_id=True --gpu_id=-1 --max_seq_len=8192 \
    --user_path=/opt/myDocuments/all/ --langchain_mode='UserData' --max_quality=True \
    --add_chat_history_to_context=True --keep_sources_in_context=True --enable_ocr=True --enable_doctr=True \
    --answer_with_sources=True --show_link_in_sources=True --append_sources_to_chat=True \
    --hf_embedding_model="intfloat/multilingual-e5-large-instruct" \
    --memory_restriction_level=0 --score_model=None --verbose=True --debug=True \
    --show_examples=True --compile_model=True \
    --share=True
``` 
and I get the following error:
```shell
FileNotFoundError: [Errno 2] No such file or directory: '/root/.cache/torch/sentence_transformers/intfloat_multilingual-e5-large-instruct/sentence_xlnet_config.json'
```

I have used the repository with the last commit at the time I write this issue:
- Commit: 47e14672
- SO: Ubuntu 20.04

As you can see the problem point out that problem is from the embedding model "intfloat/multilingual-e5-large-instruct", with the embedding model ""intfloat/multilingual-e5-large" I did not get that kind of error.

Thanks in advance for help.

llmwesee:
I'm also facing the same thing when running docker offline

pseudotensor:
If I'm online, I see the same error.

```
(h2ogpt) jon@pseudotensor:~/h2ogpt$ python generate.py     --base_model=llama --use_gpu_id=True --gpu_id=-1 --max_seq_len=8192     --user_path=user_path --langchain_mode='UserData' --max_quality=True     --add_chat_history_to_context=True --keep_sources_in_context=True --enable_ocr=True --enable_doctr=True     --answer_with_sources=True --show_link_in_sources=True --append_sources_to_chat=True     --hf_embedding_model="intfloat/multilingual-e5-large-instruct"     --memory_restriction_level=0 --score_model=None --verbose=True --debug=True     --show_examples=True --compile_model=True     --share=True
```
gives:
```
[...]

PR #1775, opened 2024-07-30
Instructor-large no longer supported for a while by sentence_transformers, need to support newer embedding models

pseudotensor:
Fixes https://github.com/h2oai/h2ogpt/issues/1774


pseudotensor:
If this causes anyone issues, just do the below to get back instructor-large support for existing databases:

```
pip install sentence_transformers==2.2.2
```

unable to support both old and new at same time.

PR #1828, opened 2024-09-05
Expose agents port

achraf-mer:


helm/h2ogpt-chart/templates/service.yaml
```diff
@@ -16,6 +16,9 @@ spec:
     - protocol: TCP
       port: {{ .Values.h2ogpt.service.webPort }}
       targetPort: 7860
+    - protocol: TCP
```
EshamAaqib, line 19: If having multiple ports we will need to add names to it otherwise might fail

EshamAaqib: similar to in https://github.com/h2oai/h2ogpt/pull/1829/files

pseudotensor (approved):

PR #1836, opened 2024-09-10
Multi agent type support

fatihozturkh2o:
image
* More about AutoGen GroupChat: https://microsoft.github.io/autogen/docs/tutorial/conversation-patterns#group-chat

- Introduces new agent flow along with new agents (see the diagram above)
   - New agent flow can be the base for the future specialized agent orchestration
   - In the current flow, all the 'If' conditions and checks are piled up in CodeWriterAgent system prompt and weaker models easily struggle with this a lot of ifs.
   - With this flow, agents are aware of each others' presence and what they are capable.
   - Flexible to introduce new agents.
   - MainGroupChatManager decides when to select the next agent until the user request is met.
      - Or direct control over which agent to talk (in case it's needed, the manager can pick next agent via user instruction) 
   - For now, as an alternative to Code Group, there is only Chat Agent that is only capable of simple chatting. If you ask questions like 'who are you', 'give me a joke' only that agent is expected to be picked up instead of going through code generation and execution steps.
   - For multi_agent type, I'm also bringing customized 'chat_history summarization'. As an example, if it's only chat_agent answer in the chat_history, then the last message will be the summary. If there were multiple agent interactions in the chat_history, then summarizing the process via LLM with a customized summarization prompt of mine.
  - Also added one more instruction in code_writer_agent system prompting to solve the following: In my tests, realized that sometimes, after code execution is completed, code writer agent wraps code output with ```output```, and hence, code execution agent thinks there is a new code block and tries to execute it and fails. Later on, this runs into a loop.


pseudotensor:
looking great!

openai_server/agent_prompting.py
```diff
@@ -190,6 +190,7 @@ def agent_system_prompt(agent_code_writer_system_message, agent_system_site_pack
   * Always use full sentences, include all items in any lists, etc.
   * i.e. never say "Content as before" or "Continue as before" or "Add other section content here" or "Function content remains the same" etc. as this will fail to work.
   * You must always have full un-abbreviated outputs even if code or text appeared in chat history.
+  * If you are just putting some code execution output in your response, dont wrap the output with ```output``` because otherwise code_executor_agent thinks that it's a code block to be executed. Find another way to display code output with markdown syntax.
```
pseudotensor, line 193: Not sure this is relevant for last week.  I only execute code that has the # execution: true tag.  Normal markdown is ok.

openai_server/autogen_2agent_backend.py
```diff
@@ -72,11 +72,17 @@ def run_autogen_2agent(query=None,
     # iostream = IOStream.get_default()
     # iostream.print("\033[32m", end="")
 
-    from openai_server.autogen_agents import get_execution_agent
-    code_executor_agent, executor = \
-        get_execution_agent(autogen_run_code_in_docker, autogen_timeout, agent_system_site_packages,
-                            autogen_max_consecutive_auto_reply, autogen_code_restrictions_level,
-                            agent_venv_dir, temp_dir)
+    from openai_server.autogen_agents import get_code_execution_agent
```
pseudotensor, line 75: I'm not sure this refactor is useful.  It's not saving much, likely best to just keep separate in each agent code.

openai_server/autogen_agents.py
```diff
@@ -1,49 +1,11 @@
...
-        else:
-            from autogen.coding.local_commandline_code_executor import \
-                LocalCommandLineCodeExecutor as H2OLocalCommandLineCodeExecutor
-        executor = H2OLocalCommandLineCodeExecutor(
-            timeout=autogen_timeout,  # Timeout for each code execution in seconds.
-            virtual_env_context=virtual_env_context,
-            work_dir=temp_dir,  # Use the temporary directory to store the code files.
-        )
-
+def get_code_execution_agent(
```
pseudotensor, line 5: Do we use the executor separately?  Please explain why this separation is useful, I couldn't tell.

openai_server/autogen_multi_agent_backend.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+
+    # A little sumamry clean-up
+    summary = summary.replace("ENDOFTURN", " ").replace("TERMINATE", " ")
+    # Update chat_result with summary
+    chat_result.summary = summary
+
+    # TODO: put fake token cost for now
+    # for chats containing GroupChatManager, costs seem to be not calculated properly
+    chat_result.cost={
```
pseudotensor, line 192: Note that h2oGPTe doesn't consume this cost, only tokens (which should already be ok for 2 agent case), and it determines it's own cost.

pseudotensor (commented):
let's remove t he prompt line and I'll refactor the autogen_code_restrictions_level stuff so it's clear it only applies to guardrail part.

openai_server/agent_prompting.py
fatihozturkh2o: as discussed in slack, removed: https://github.com/h2oai/h2ogpt/pull/1836/commits/d37340d5be0a6321ef21d6d44bee3daadec28e2e

openai_server/autogen_multi_agent_backend.py
fatihozturkh2o: yeah I'll do more research/test for this part and see why num tokens do not return.

pseudotensor:
Want to merge so we can move on and add my changes we discussed.  Let's make new issue for the cost issue.

PR #1846, opened 2024-09-16
new multi agent  flow with terminate agent

fatihozturkh2o:
image

- Better for prompts where user is only asking for the code but not for the execution.
   - Now GroupManager decides whether code execution is needed or not based on # execution: VALUE tag coming from code generator agent.
- Now Group manager has capability of terminating the chat as soon as user request is met. It's done by selecting the next agent as 'terminate_agent'.
   - This termination process can be extended to detecting repetitive errors in the agents' conversation. If Group manager decides that there has been repetitive/unsolvable errors in the chat history, it can decide to terminate the chat.

Based on current h2ogpte agents pytest, it shows similar results to the other agent types. The most distinctive success comes with the question "Write a code to show an image in matplotlib.  Do not execute any code." Because this new flow is the only flow that does not even call 'code_execution_agent'.

- [ ] testing

### DIFF
--- a/openai_server/autogen_agents.py
+++ b/openai_server/autogen_agents.py
@@ -30,6 +30,21 @@ def get_code_writer_agent(
         code_writer_system_prompt:str | None = None,
         autogen_max_consecutive_auto_reply:int = 1,
         ):
+    
+    code_writer_system_prompt += (
+        "\n\n"
+        "* Important: If you have past conversation context, "
+        "and if you see code execution errors, "
+        "you should focus on fixing the errors in the code. "
+        "* Important: If you realize you run into a lot of similar "
+        "errors for a certain approach, then instead of trying to fix the errors, "
+        "you should try a different approach. "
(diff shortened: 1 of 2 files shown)


pseudotensor:
`Write a code to show an image in matplotlib. Do not execute any code.` works normally for the 2 agent case.

fatihozturkh2o:
> `Write a code to show an image in matplotlib. Do not execute any code.` works normally for the 2 agent case. It does not execute the code.

Yeah, but, with 2agent, the code execution agent is called anyways like in this example: (even code generation is called once more though it's not needed)

#### 2agent
https://github.com/user-attachments/assets/48ee5cd5-8533-40ef-a6c4-8052f7c41e0f


I'm aiming not even calling code_executor agent with the new flow (because # execution: false information is out there anyway, thought the manager could make use of it during agent selection)

#### PR multiagent
https://github.com/user-attachments/assets/7276743c-89db-4c8d-b4f7-611e46706994

PR #1847, opened 2024-09-17
More tools for Agents

fatihozturkh2o:
- Adds Image generation tool for agents
- Adds STT (Audio transcription) tool for agents


openai_server/agent_tools/image_generation.py
```diff
@@ -0,0 +1,54 @@
...
+        if not os.path.exists(base_path):
+            os.makedirs(base_path)
+        full_path = os.path.join(base_path, args.file_name)
+
+        # Write the image data to a file
+        with open(full_path, "wb") as img_file:
+            img_file.write(image_data)
+
+        print(f"Image successfully saved to the path: {full_path}")
+    except Exception as e:
```
pseudotensor, line 50: Why is this exception handle required?  Shouldn't be.

openai_server/server.py
```diff
@@ -720,6 +720,7 @@ async def handle_image_generation(request: Request):
         n = body.get('n', 1)  # ignore the batch limits of max 10
         response_format = body.get('response_format', 'b64_json')  # or url
 
+        # TODO: Why not using image_request? size, quality and stuff?
```
pseudotensor, line 723: Yes we can, just didn't focus on it.

openai_server/agent_tools/image_generation.py
```diff
@@ -0,0 +1,54 @@
...
+        parser = argparse.ArgumentParser(description="Generate images from text prompts")
+        # Prompt
+        parser.add_argument("--prompt", type=str, required=True, help="User prompt")
+        # Model
+        parser.add_argument("--model", type=str, default="sdxl_turbo", help="Model name")
+        # File name
+        parser.add_argument("--file_name", type=str, default="output.jpg", help="Name of the output file")
+        args = parser.parse_args()
+        ##
+        base_url = os.getenv('H2OGPT_OPENAI_BASE_URL')
```
pseudotensor, line 18: Needs to be separate envs for the OpenAI endpoint so can point to anything, not just our own OpenAI proxy.  E.g. IMAGEGEN_OPENAI_BASE_URL.  Same for the key name.

These key names need to be then added to the output guardrails.

In enterprise, we can set it in helm chart and docker compose to point to gpt.h2o.ai with the appropriate key.

openai_server/agent_prompting.py
```diff
@@ -501,13 +501,32 @@ def get_mermaid_renderer_helper():
     return mmdc
 
 
+def get_image_generation_helper():
+    # TODO: how to check existing visible_image_models here and assign new env variable here before calling image_generation.py?
+    # Or always use flux-schenell for now?
+    cwd = os.path.abspath(os.getcwd())
+    base_path = os.getenv("H2OGPT_OPENAI_BASE_FILE_PATH", "./openai_files/")
```
pseudotensor, line 508: Needs to be conditionally added upon non-empty value of IMAGEGEN_OPENAI_BASE_URL

openai_server/agent_tools/image_generation.py
fatihozturkh2o: Ive introduced IMAGEGEN_OPENAI_BASE_URL: https://github.com/h2oai/h2ogpt/pull/1847/commits/16fcdc34f970d33a9707ba89eaa0ccd20dae47a1

> These key names need to be then added to the output guardrails.
Can you elaborate this a bit more? Does it mean we need to update 'sanitize_command' of ours? If so how exactly do you think? Like should we disable 'displaying' these keys, or disable assigning new values to these keys?

openai_server/agent_prompting.py
fatihozturkh2o: Updated accordingly: https://github.com/h2oai/h2ogpt/commit/16fcdc34f970d33a9707ba89eaa0ccd20dae47a1

openai_server/agent_tools/image_generation.py
fatihozturkh2o: I thought there could be things that might go wrong, so by printing it out I could let agents know about it 🤔  But maybe I'm already handling all the scenarios via asserts. So removed: https://github.com/h2oai/h2ogpt/pull/1847/commits/4ffb389ee06e5af1a412f44381066d3efad9eef4

pseudotensor: If any python tool fails it's already handled by code execution handling, no extra asserts needed for that aspect.  No hard failures occur just because a code fails to run.

openai_server/server.py
pseudotensor: Feel free to update the tool and OpenAI image generation code to better match OpenAIs for these things.  Especially quality is good one, so can ask for rough image and be quick with fewer inference steps.

We can also expose in extra_body other things not normally possible via OpenAI, like guidance_scale, max_sequence_length.

fatihozturkh2o: > Feel free to update the tool and OpenAI image generation code to better match OpenAIs for these things.

I'm working on this for the past hours but couldnt make it work.  I simply just wanted to iclude/pass 'size' parameter here like this:
```
body_image = dict(prompt=prompt, langchain_action='ImageGen', visible_image_models=model, imagegen_size=size)
    response = completions(body_image)
```

However, even though I introduce new 'imagegen_size' parameter in gen/main and gen/evaluate I can not get to see what I'm passing via 

```
client.images.generate(
        prompt=prompt,
        model="sdxl_turbo", # flux.1-schnell, sdxl_turbo
        size=700,
        )
```

Based on my debugs, it seems during gradico client.predict this new parameter is kinda omitted but I can not point where this is happenning actually.

So I need your help here to learn more about how this flow works.

pseudotensor: ok looking

pseudotensor: pushed changes.  See if you can update the tool to use the new options.

pseudotensor (approved):

PR #1854, opened 2024-09-25
Agentic improvements

fatihozturkh2o:
Introduces some improvements for the agentic flows:
- Rename `TERMINATE` command to` <FINISHED_ALL_TASKS>` everywhere relevant
- Renaming `image_query` tool to `ask_question_about_image` in order to make it more explicit so that models won't confuse image_query with other tasks like querying for online images etc.
- Introducing new `download_one_web_image` tool, that especially helps weak models with tasks involving subtasks like downloading online images. Otherwise, they seem to struggle big time until finding the right method.
- Fixing `wolfram_query` code example where we didnt put '--prompt' parameter and the models were running into error in their first attempts.
- Adding `# execution : true ` into all the example codes we are providing in the system prompt. It seems some of them dont have this mark in it, so some OSS models follow it as is without proper marks. Examples without the fix ⬇️ 
   - Example code from open source models (this one from llama_3.2_90B):
   - image
   - Another example from the same model:
   - image

- Introducing new small `code_writer_terminate_func` for code_writer_agent, so that whenever `code_executor_agent` returns empty strings `code_writer_agent` can also end the conversation.
   - If you just remove terminate_func from code_writer_agent, then it goes into unnecessary loops when `code_execution` returns empty strings ''. Code_writer_agent keeps talking..
- Some wording improvements for `<no_code_executed_notes>` message.


openai_server/agent_prompting.py
```diff
@@ -589,6 +589,59 @@ def get_audio_transcription_helper():
...
+python {cwd}/openai_server/agent_tools/image_downloader.py --text "Text to search for" --file "file_name.jpg"
+```
+* usage: python {cwd}/openai_server/agent_tools/image_downloader.py [-h] --text "TEXT TO SEARCH FOR" --file "FILE_NAME"
+* The image_downloader tool uses the Google Search API to download images from the web.
+* The image_downloader tool can be used to download images for use in generating responses or for analysis.
+* The image_downloader tool has to be your first option for downloading images from the web.
+"""
+    return image_download
+
+def get_final_system_highlights():
```
pseudotensor, line 613: Seems to be redundant with existing parts of the prompt.

openai_server/agent_prompting.py
```diff
@@ -589,6 +589,59 @@ def get_audio_transcription_helper():
         audio_transcription = ''
     return audio_transcription
 
+def get_image_download_helper():
```
pseudotensor, line 592: A risk with this kind of tool is it overlaps conceptually with general tasks.  Especially the name implies some general tool, and even sonnet3.5 can misuse tools based upon their name.  When we add tools, we have to be sure it has very little overlap with any other types of tasks.  Also, just because we see a failure mode, I think it's a bit overfit to add a tool to cover, especially when the tool code is relatively specific use case.  By using SERP, it's super specific and is trying to cover a very specific failure (I guess) for a very general task.

fatihozturkh2o: renamed

openai_server/agent_prompting.py
fatihozturkh2o: removed get_final_system_highlights. After playing with OSS models more, realised that the main issue was coming from our example prompts in which we didnt put these execution marks, so OSS models were copying what we provided. After fixing them, the issue seems to be gone and I dont see a value with this function. See more in the PR description.

openai_server/agent_prompting.py
```diff
@@ -615,6 +669,8 @@ def get_full_system_prompt(agent_code_writer_system_message, agent_system_site_p
 
     agent_tools_note = f"\nDo not hallucinate agent_tools tools. The only files in the {path_agent_tools} directory are as follows: {list_dir}\n"
 
-    system_message = agent_code_writer_system_message + image_query_helper + mermaid_renderer_helper + image_generation_helper + audio_transcription_helper + agent_tools_note + chat_doc_query
-    # TODO: Also return image_generation_helper and audio_transcription_helper ? 
+    system_message = agent_code_writer_system_message + image_query_helper + mermaid_renderer_helper + image_generation_helper + audio_transcription_helper + image_download_helper + agent_tools_note + chat_doc_query
+    
+    # Emphasize the most important points at the end
+    system_message += get_final_system_highlights()
```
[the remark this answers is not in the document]
fatihozturkh2o: get_final_system_highlights is removed

openai_server/agent_prompting.py
```diff
@@ -90,6 +93,7 @@ def agent_system_prompt(agent_code_writer_system_message, agent_system_site_pack
 <query_understanding>
 * If the user directs you to do something (e.g. make a plot), then do it via code generation.
 * If the user asks a question requiring math or puzzles, then solve it via code generation.
+* If the user asks a question that requires even simple math operations like addition or counting, always solve it via code generation because you cannot do math by heart.
```
pseudotensor, line 96: Seems too similar to previous line.  I'd expect rewording/improving the previous line should be ok.  I had a longer line, but I wanted it to be even more basic so gpt-4o would not try to solve things.  So e.g. we can make the first line say: `If the user asks a question requiring any kind of math or puzzles, then you MUST solve it via code generation.`

openai_server/agent_prompting.py
```diff
@@ -453,7 +457,7 @@ def get_standard_answer(prompt, text_context_list, image_file=None, chat_convers
     return text
 
 
-def get_image_query_helper(base_url, api_key, model):
+def get_ask_question_about_image_helper(base_url, api_key, model):
```
pseudotensor, line 460: global thing.  Change `image_query_helper` to `ask_question_about_image_helper` -- there's some mixed changes currently.

openai_server/agent_utils.py
```diff
@@ -272,7 +272,7 @@ def get_ret_dict_and_handle_files(chat_result, temp_dir, agent_verbose, internal
             extracted_summary = extract_xml_tags(chat_result.summary, tags=['constrained_output'])['constrained_output']
             if extracted_summary:
                 chat_result.summary = extracted_summary
-        chat_result.summary = chat_result.summary.replace('ENDOFTURN', '').replace('TERMINATE', '')
+        chat_result.summary = chat_result.summary.replace('ENDOFTURN', '').replace('<FINISHED_ALL_TASKS>', '')
```
pseudotensor, line 275: Some missing changes of TERMINATE if you search through code base.  E.g.
```
summary = summary.replace("ENDOFTURN", " ").replace("TERMINATE", " ")
```

and a 3 in comments mention `TERMINATE` still.

image

pseudotensor: Note that some TERMINATE occurrences I didn't mention above are from autogen code copy or tests that happen to have it in it.  Those do not need changing.

image

openai_server/autogen_utils.py
```diff
@@ -365,8 +365,12 @@ def _execute_code_dont_check_setup(self, code_blocks: List[CodeBlock]) -> Comman
 * This is expected if you had code blocks but they were not meant for python or shell execution.
 * For example, you may have shown code for demonstration purposes.
 * If you intended to execute code, be sure to add the comment: # execution: true and try again.
-* Otherwise, if no code execution was expected, then do not respond or react to this "no_code_execution" text and instead just directly and immediately summarize the actual answer to the user's original question and then TERMINATE.
-* Do not thank the user for telling you that the code was not executed.
+* Otherwise, if no code execution was expected:
```
pseudotensor, line 368: We need better prompting here.  That is, we cannot expect a normal model to understand an indent and then new sentences very well.  IMO every single sentence should try to stand alone.  If that single sentence is somehow opposite, it won't work well.

E.g. `Instead just directly and immediately provide the actual answer to the user's original question.` does not stand alone.

openai_server/agent_prompting.py
fatihozturkh2o: thanks for the catch, done: https://github.com/h2oai/h2ogpt/pull/1854/commits/e56f2cd22f98d44ed2ab9b1339630baabb127238

openai_server/agent_prompting.py
fatihozturkh2o: How about this one: https://github.com/h2oai/h2ogpt/pull/1854/commits/464aa281fa8cd087d17fc6f7063320b4a0895829 ?

I see value highlighting 'even simple operations like addition our counting' instead of just saying 'requires math...'.

pseudotensor: ok good

pseudotensor (approved):

PR #1863, opened 2024-10-01
Final response improvements

fatihozturkh2o:
- Improving final agent response via passing instructions in code execution notes.
   - After executed_code_note, now also passing final_answer_guidelines like this:
```
        ret = self.truncate_output(ret)
        ret = self.executed_code_note(ret)
        ret = self.final_answer_guidelines(ret)
```
With the new instructions, aiming: 
- to make final message sound directly talking to the main user
- to have the actual answer the user was looking for (e.g. the code itself rather than the ouput etc.)
- to avoid internal agent expressions like 'Thank you for providing code execution output' etc.
- to still keep the LLM's own expressiveness after providing the actual answer (You can see it in the recording below, claude now starts with the actual answer but continues with its own explanations)

#### How it looks on main (without final_answer_guidelines)

https://github.com/user-attachments/assets/9402dda7-df4e-44da-b446-3172754b5905

#### How it looks on this PR (with final_answer_guidelines)

https://github.com/user-attachments/assets/86012a57-c5c6-4a45-97ad-b86170dd7e2b

**Personal Tests**
- Sonnet 3.5 never failed what I was looking for based on my personal tests. (Always provides the code in the final answer now, or never Thanks etc.)
- llama 405B also shows similar performance to sonnet 3.5 in terms of following new instructions
- llama 3.2 90b also shows visible improvement in terms of following the new instructions

Comparing to the PR: https://github.com/h2oai/h2ogpt/pull/1861
- This PR has two main advantages: 1. No extra llm call, 2. No interruption with the LLM's own expressiveness by not forcing llm to summarize the history but rather paying attention to final answer tone and wordings.


fatihozturkh2o:
RAG question on main branch:
Screenshot 2024-10-01 at 4 30 31 PM

same RAG question on this PR branch:
Screenshot 2024-10-01 at 4 30 42 PM

openai_server/autogen_utils.py
```diff
@@ -417,6 +417,25 @@ def executed_code_note(self, ret: CommandLineCodeResult) -> CommandLineCodeResul
...
+* If there is not enough information to provide a direct answer, mention that you couldn't find enough information for the task or for some of the sub-tasks.
+* If there were any crucial internal steps or discoveries in the chat history, mention them briefly as well if they are directly related to the answer.
+* If the user was asking for seeing or having codes directly, make sure to provide the code block in the final answer because the user waiting for it.
+* If the user was asking for images or there were images generated in the chat history, you should have them by adding inline markdowns of any key image, chart, or graphic. Just use '![image](filename.png)' without any code block and only use the basename of the file, not the full path. Also, pay attention to placing images in relevant places in the final answer.
+* You can use markdown syntax for formatting the text (e.g. long lists, tables, headers etc.) in the response to make it more readable and easy to follow.
+* Terminate the chat by having <FINISHED_ALL_TASKS> string in your final answer.
```
pseudotensor, line 437: This whole text sounds like it applies to the very end, but this text block would appear after every code execution event.  I'd expect all this phrasing will lead to early termination of the iterations.

pseudotensor: These kinds of phrasings are more appropriate for the no code block, because that is when there is a high likelihood of it being the final answer, and then it's job is to just repeat that final answer.

openai_server/autogen_utils.py
```diff
@@ -417,6 +417,25 @@ def executed_code_note(self, ret: CommandLineCodeResult) -> CommandLineCodeResul
...
 </code_executed_notes>
+"""
+        return ret
+
+    def final_answer_guidelines(self, ret: CommandLineCodeResult) -> CommandLineCodeResult:
+        if ret.exit_code == 0:
+            ret.output += """
+\nIf there is no more task left, you should terminate the chat with your final answer.
+<final_answer_guidelines>
+* Identify the first user request in the chat history and start your response by providing a direct answer to that request by only using the information avaiable in the chat history.
```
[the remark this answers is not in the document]
fatihozturkh2o: Even though I haven't noticed it in my test chats, it sounds reasonable what you said. So changed. Wdyt now? https://github.com/h2oai/h2ogpt/pull/1863/commits/aadf97b3de785970a68cbf2978f4fdfd1c7de1cb

pseudotensor (approved):

pseudotensor (changes requested):
see comments

pseudotensor:
Yes let's see if we can isolate the "final" block and (without repeating code) use in both system prompt and that final block.

pseudotensor:
@fatihozturkh2o 

Did you plan to still work on this?  I think the resulting conclusion about the below is fine.

> Yes let's see if we can isolate the "final" block and (without repeating code) use in both system prompt and that final block.

pseudotensor (approved):

PR #1868, opened 2024-10-03
code_block and ENDOFTURN control via API streaming

fatihozturkh2o:
This PR:
- makes sure that each code_writer_agent message will end with ENDOFTURN string
- makes sure that code_writer_agent string will end after each executable code block to be able to run it (num code block limit is 1)

@pseudotensor this is my first attempt based on discussion on slack. Here I'm controlling number of code blocks + ENDOFTURN behaviours on API steaming side.

There were things that I could do better but didnt know how, so left some TODO comments. Feel free to change the code and let me know what you think with the code.

### DIFF
--- a/openai_server/backend.py
+++ b/openai_server/backend.py
@@ -274,26 +274,22 @@ async def get_response(chunk_response=True, **kwargs):
     client = get_client(user=kwargs.get('user'))
 
     res_dict = {}
-    def check_if_response_was_from_code_writer_agent() -> bool:
+
+    def check_if_executable_code_block_hit_limit(response: str, num_limit=1) -> bool:
         # TODO: Need a better way to check if the request is coming from code_writer_agent or not
         # Maybe new attribute in extra_body called 'agent_name'?
         # Below method is a hacky way to check if the request is coming from code_wrigter_agent for now.
         system_prompt = kwargs.get('system_prompt', '')
-        if system_prompt.startswith(
+        if not system_prompt.startswith(
             'You are a helpful AI assistant.  Solve tasks using your coding and language skills'
         ):
             # This is a request coming from code_writer_agent.
-            return True
-        return False
-
-    def check_if_executable_code_block_hit_limit(response: str, num_limit=1) -> bool:
-        if not check_if_response_was_from_code_writer_agent():
             return False
-        from autogen.code_utils import extract_code
+        from autogen.coding import MarkdownCodeExtractor
         # TODO: num_limit might be a parameter in extra_body ?
-        extracted_code_blocks = extract_code(response, pattern=r"```[ \t]*(\w+)?[ \t]*\r?\n(.*?)\r?\n[ \t]*```")
-        code_blocks = [block[1] for block in extracted_code_blocks if block[0] not in ['', 'unknown']]
-        executable_code_blocks = [block for block in code_blocks if '# execution: true' in block]
+        extracted_code_blocks = MarkdownCodeExtractor().extract_code_blocks(response)
+        code_blocks = [block.code for block in extracted_code_blocks if block.language not in ['', 'unknown']]
+        executable_code_blocks = [code for code in code_blocks if '# execution: true' in code]
         return len(executable_code_blocks) >= num_limit
 
 
--- a/openai_server/backend.py
+++ b/openai_server/backend.py
@@ -274,6 +274,28 @@ async def get_response(chunk_response=True, **kwargs):
     client = get_client(user=kwargs.get('user'))
 
     res_dict = {}
+    def check_if_response_was_from_code_writer_agent() -> bool:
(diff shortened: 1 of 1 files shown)


openai_server/backend.py
```diff
@@ -289,11 +311,20 @@ async def get_response(chunk_response=True, **kwargs):
...
+                    break
                 last_response = response
                 await asyncio.sleep(0.005)
             await asyncio.sleep(0.005)
             job_outputs_num += job_outputs_num_new
 
+        if check_if_response_was_from_code_writer_agent():
+            # If the last response was from code_writer_agent, then we always have to return "ENDOFTURN" at the end of the message.
+            print("Finishing code_writer_agent message with 'ENDOFTURN'", flush=True)
+            yield "\n\nENDOFTURN"
```
pseudotensor, line 326: Should already be taken care of here: https://github.com/h2oai/h2ogpt/blob/3828ca7d2ed117e3b33d98a939610bf18619b950/openai_server/autogen_utils.py#L746

openai_server/backend.py
```diff
@@ -274,6 +274,28 @@ async def get_response(chunk_response=True, **kwargs):
...
+            # This is a request coming from code_writer_agent.
+            return True
+        return False
+
+    def check_if_executable_code_block_hit_limit(response: str, num_limit=1) -> bool:
+        if not check_if_response_was_from_code_writer_agent():
+            return False
+        from autogen.code_utils import extract_code
+        # TODO: num_limit might be a parameter in extra_body ?
+        extracted_code_blocks = extract_code(response, pattern=r"```[ \t]*(\w+)?[ \t]*\r?\n(.*?)\r?\n[ \t]*```")
```
pseudotensor, line 294: This should use the same code that exists for detecting and extracting code blocks to avoid divergent behavior.

openai_server/backend.py
```diff
@@ -274,6 +274,28 @@ async def get_response(chunk_response=True, **kwargs):
     client = get_client(user=kwargs.get('user'))
 
     res_dict = {}
+    def check_if_response_was_from_code_writer_agent() -> bool:
+        # TODO: Need a better way to check if the request is coming from code_writer_agent or not
+        # Maybe new attribute in extra_body called 'agent_name'?
+        # Below method is a hacky way to check if the request is coming from code_wrigter_agent for now.
+        system_prompt = kwargs.get('system_prompt', '')
+        if system_prompt.startswith(
+            'You are a helpful AI assistant.  Solve tasks using your coding and language skills'
```
pseudotensor, line 283: This is too hacky, but use_agent variable can be used.

openai_server/backend.py
```diff
@@ -289,11 +311,20 @@ async def get_response(chunk_response=True, **kwargs):
                         yield chunk
                 else:
                     yield response
+                if check_if_executable_code_block_hit_limit(response):
+                    print("Hit executable code_block limit for code_writer_agent, stopping streaming", flush=True)
+                    job.cancel()
```
pseudotensor, line 316: Does this actually work to cancel the upstream job? In past gradio jobs that are cancelled have no effect on actual inner threads, they just stop the generation consumption, but the inner thread doing work to talk to vllm etc. will continue.  So it's not super helpful.  We should check what backend is doing.

openai_server/backend.py
fatihozturkh2o: Hmm, ok then it needs to be investigated why it didnt work in your example 🤔

openai_server/backend.py
fatihozturkh2o: yeah know it's hacky for now, but I don't think use_agent would be the right way though since at this point it's false 🤔  That's why I was thinking of a new 'agent_name' variable maybe?

openai_server/backend.py
fatihozturkh2o: now using the same code extractor in the autogen local executer:
https://github.com/microsoft/autogen/blob/180c17c1f4b292f798d055fcd02a3f6b9e94d2ab/autogen/coding/local_commandline_code_executor.py#L182

change: https://github.com/h2oai/h2ogpt/pull/1868/commits/c76e0350eb8ed397eaa14ca404794880c0a80830

openai_server/backend.py
pseudotensor: It should only be false for non-agent case, e.g. what the agent talks to.

openai_server/backend.py
pseudotensor: We could add more logging, but one can try --verbose=True when starting up.

openai_server/backend.py
pseudotensor: Ya, I don't understand the issue either, needs to be invesigated.

fatihozturkh2o:
handled on code_executer side for now: https://github.com/h2oai/h2ogpt/pull/1877

PR #1878, opened 2024-10-17
final answer improvements

fatihozturkh2o:
- visiting final_answer_guidelines when there is no code but FINISHED_ALL_TASKS tag too.
- new guideline point for llm to start its final answer by focusing on user's initial request.


pseudotensor (approved):
Looks good, let's see how GAIA runs.

PR #1882, opened 2024-10-22
Fix: Correct URL Scheme for File Links in HTML Output

Sivakajan-tech:
Windows often requires file paths to be formatted differently. The path used is formatted correctly for Window. Using file:/// for better cross-platform support:

```
return """<font size="%s"><a href="file:///%s" target="_blank"  rel="noopener noreferrer">%s</a></font>""" % (
```

Note the three slashes (file:///) in the link. This makes the link an absolute path which can be more reliable across different platforms.


pseudotensor (approved):
thanks!

PR #1885, opened 2024-10-24
tracking agent_tool usage counts

fatihozturkh2o:
Tracking agent_tool usage counts. These counts can be used for future tool-based limitations for the agents

Example usage looks like: `{'image_generation.py': 2, 'ask_question_about_image.py': 1}
`


openai_server/autogen_utils.py
```diff
@@ -430,10 +431,24 @@ def _execute_code_dont_check_setup(self, code_blocks: List[CodeBlock]) -> Comman
                 ret = CommandLineCodeResult(exit_code=1, output=str(e))
             else:
                 raise
+        
+        # Update agent tool usage if there is any
+        self.update_agent_tool_usages(code_blocks)
```
pseudotensor, line 436: Should only be code_block_exec right?

pseudotensor: updated that

fatihozturkh2o: ah yes, good catch

PR #1916, opened 2025-01-03
Fixing small typos in `agent_system_prompt`

svilupp:
Thanks for open-sourcing! It was a great read.

When going through the system prompt, I noticed some small typos, so I opened a PR:
- there were typos in words (packages, a lot)
- extra words (`should be include`)
- missing punctuation
- ? double spaces within bullets. I noticed too late that you have 2000 occurrences of those, so I should probably revert this change?

I didn't fix all grammar issues, just the big things that stood out as confusing. If it confuses me, it would confuse an LLM too.

Feel free to ignore/close if not helpful.


pseudotensor (approved):
thanks!

